PR #27, opened 2017-02-19
Added lastIndex utility property

MrPlow442:
Kotlin std array and list extensions have this property and it's super useful in for loops


czyzby:
That's a great idea, thanks! I made this property null-safe (you can safely use it on a null `Array` and it will return -1) and added it to the "primitive" LibGDX collections (`IntArray`, `FloatArray`, `BooleanArray`). These should already be available in the latest `1.9.5-SNAPSHOT`.

PR #37, opened 2017-03-30
add wrappers around applicationlistener, inputprocessor, inputlistener

sreich:
first batch of my code moving upstream :tada: 

could you guide me a bit on how to test the interfaces? couldn't figure out how from looking at the other code.

regarding scene2d, are scene2d and visui modules mutually exclusive? if yes, that means that same code(ktxinputlistener) should reside in both, yeah?


czyzby:
> could you guide me a bit on how to test the interfaces? couldn't figure out how from looking at the other code.

You don't. You test only the logic of your libraries/applications, "testing" interfaces is almost like testing the language itself.

There are two cases I can think of such tests could be sort of useful:
- When _extending_ third-party interfaces to implement all methods with empty blocks, create a class with no custom methods implementing your wrapper interface. If it fails to compile, the third-party interface was extended. This could work for `ApplicationListener`, for example (although I doubt that they will ever change this interface, breaking hundreds of applications, but still).
- When providing defaults for an interface that requires you to return some values. You can basically check if your values returned by your default methods make sense. `InputProcessor` could benefit from this.

Again, I'm not telling you to write these, just saying that there are cases that they _would_ make sense. Sorta. ; )

> regarding scene2d, are scene2d and visui modules mutually exclusive? if yes, that means that same code(ktxinputlistener) should reside in both, yeah?

Yes, they are. Even their building blocks work differently, handling `Cell` and `Node` instances with different approaches. `scene2d` building blocks are inlined, `vis` are not. Before the final non-beta release, I plan to choose the better approach and make both of these modules compatible - `vis` should depend on and extend `scene2d`.

Anyway, both of these libraries are mostly for the UI _building_ phase. General Scene2D utilities go to `ktx-actors`. I think you should move `KtxInputListener` there, to `events.kt` file.

czyzby:
Oh, and please add info about these interfaces in changelog and `README` files.

sreich:
Yeah it was mostly the compile guarantees I wanted. I like the thought that if something gets deleted or renamed here, it flags it.

> Before the final non-beta release, I plan to choose the better approach and make both of these modules compatible - vis should depend on and extend scene2d.

Sounds like you should make an issue for that? 😉

czyzby:
> I like the thought that if something gets deleted or renamed here, it flags it.

If something gets deleted or renamed, the interface itself won't compile (since `override` is a keyword), so additional tests are not required. However, to ensure that no new methods are added unnoticed, you can just add a simple class like `class ShouldImplementAllMethodsWithDefaults : MyInterface` in tests package. If a new method is added, test sources will fail to compile.

> Sounds like you should make an issue for that?

Still not sure which approach is better. I like the inlining that the builders from `scene2d` provide, but since you cannot add a default functional interface implementation in inline methods, you basically have to create 2 methods for each added widget. `vis` approach might be a little more generic, at least in terms of the implementation itself.

It comes down to how you treat the additional data like `Cell` or `Tree.Node`. `scene2d` passes it as ("optional") lambda argument in building block. `vis` returns it from the building block. So, for example, this is how you'd build a `Table` with `Label` that you need a reference to:

```Kotlin
// scene2d - notice that label is available directly from the building block.
table {
  val label = label("Text.") { cell ->
    setWrap(true)
    cell.expand()
  }
}

// vis - notice that you have to call .actor
table {
  val label = label("Text.") {
    setWrap(true)
  }.expand().actor
}

// vis approach might lead to poor readability with too many nested actors, since
// their configuration is separated from the body of their building block:
table {
  textButton() {
    textButton() {
      textButton() {
        // Again, which button is filled()?
      }.fill()
    }.expand()
  }.grow()
}
```

There's also the issue of scoping. I'm pretty sure Kotlin 1.1 solved that with builder-specific annotations - I'll have to look into that. Still, to early to create an issue, as I'm still not sure what has to be done.

sreich:
alright, new changes following discussions.

i also discovered `override fun handle(event: Event)`  in inputlistener(scene2d), wasn't implemented..i'm like 90% sure that is never null.

czyzby:
I don't think that interface "tests" need to override all methods. As I've said, `override` is a keyword in Kotlin, so basically your interface would not even compile if LibGDX changes anything. "Interface tests" can be added only to guarantee that all methods are optional and you can just implement the interfaces with an "empty" class. By implementing all methods, you're basically duplicating code for no apparent benefit. I'll change this.

sreich:
Okay, thanks.

czyzby:
> i also discovered `override fun handle(event: Event)` in inputlistener(scene2d), wasn't implemented..i'm like 90% sure that is never null.

It should not have been overridden. It handles dispatching events to other methods. If anything, this could be made final.

sreich:
Must have been why I didn't do so in my code ;)

czyzby:
Merged, refactored, uploaded. Thanks.

Issue #54, opened 2017-04-19
Expose KotlinApplication#timeSinceLastRender to allow interpolation/extrapolation


raincole:
`KotlinApplication`'s implementation of fixed timestep doesn't take interpolation/extrapolation between frames into account. For example, if the `fixedTimeStep` is `0.05`, but the first `Gdx.graphics.rawDeltaTime` is `0.08`, then it only calls `render(0.05)` for one time. In the other words, it renders the state of the game at 0.05s, at the 0.08s of the real world.

It's not that bad, but we can at least let the user decide whether he wants interpolation/extrapolation for a smoother result.

Since `KotlinApplication` doesn't expose `timeSinceLastRender`, I have to copy&paste its code and implement my own `ApplicationListener` this way:

```
  override final fun render() {
    timeSinceLastRender = Math.min(timeSinceLastRender + Gdx.graphics.rawDeltaTime, maxDeltaTime)
    while (timeSinceLastRender >= fixedTimeStep) {
      timeSinceLastRender -= fixedTimeStep
      clearScreen(0f, 0f, 0f, 1f)
      update(fixedTimeStep)
    }
    smoothRender(timeSinceLastRender / fixedTimeStep) // Extrapolation
  }
```

czyzby:
I think that was the exact point of `KotlinApplication`: provide consistent render calls with constant fixed time stamp value, smoothing raw delta time if necessary. If you need a specialized rendering approach, implementing `KtxApplicationAdapter` is your best bet. (It will be released in the next version pretty soon.)

If you want to make `KotlinApplication` more flexible, you can submit a pull request and I'll review your changes.

raincole:
The only thing I want is to make `timeSinceLastRender` so I can use it in my own render logic!

czyzby:
I see. The example was a bit misleading, I thought you want even more control over the `KotlinApplication` logic. I reviewed your changes.

czyzby:
If there's a use case and the setter is private (so you cannot affect the internal logic), I'm fine with that. The snapshot version with the improvement is being uploaded as we speak. Thanks for your suggestion and the pull request.

PR #55, opened 2017-04-19
Make timeSinceLastRender protected

raincole:
As issue #54 states, it's a useful information for the user. e.g. interpolation/extrapolation between frames.


czyzby:
Can you also update `CHANGELOG.md` with the information that now this variable is protected? Thanks.

We could also add `private set`, as this is usually not something you'd like to modify manually. Do you agree?

raincole:
Fair enough!

czyzby:
Thanks.

PR #64, opened 2017-05-05
Put kotlin-stdlib to compile scope

raincole:
If I'm not mistaken, one should only use `testCompile` scope for the libraries that are only useful in tests.

(Admittedly I changed this as a workaround because Intellij expected it. But after thinking twice I consider it reasonable)

This is pretty trivial so I didn't doc it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ subprojects {
     testCompile "org.mockito:mockito-core:$mockitoVersion"
     testCompile "io.kotlintest:kotlintest:$kotlinTestVersion"
     testCompile "com.nhaarman:mockito-kotlin:$kotlinMockitoVersion"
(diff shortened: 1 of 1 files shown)


czyzby:
We did it on purpose to ease usage of different Kotlin versions - KTX users have to explicitly declare dependency on a specific Kotlin standard library version and we do not want to override that. If I'm not mistaken, Gradle (for example) always chooses the newest library version that was declared either by your script or one of your dependencies. This isn't an issue as long as KTX depends on the same or an older Kotlin version - but we do not want to force the users to use a newer Kotlin version if they do not want that, for whatever reason.

KTX libraries that do not need Kotlin standard library features - and rely only on the basic language syntax - do not depend on it at all. Other libraries use include it as `provided` dependency.

PR #67, opened 2017-05-09
Replace mockito with mockito-kotlin

raincole:
`mockito-kotlin` covers the use cases of `mockito` so there is no need to depend on `mockito` any more.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,6 +42,7 @@ subprojects {
   dependencies {
     provided "com.badlogicgames.gdx:gdx:$gdxVersion"
     testCompile "junit:junit:$junitVersion"
+    testCompile "org.mockito:mockito-core:$mockitoVersion"
     testCompile "io.kotlintest:kotlintest:$kotlinTestVersion"
     testCompile "com.nhaarman:mockito-kotlin:$kotlinMockitoVersion"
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -42,7 +42,6 @@ subprojects {
   dependencies {
     provided "com.badlogicgames.gdx:gdx:$gdxVersion"
     testCompile "junit:junit:$junitVersion"
-    testCompile "org.mockito:mockito-core:$mockitoVersion"
```
czyzby: Just so you know, `kotlin-mockito` still includes this as compile dependency, so it's not like we're removing it from classpath. I think it's OK to keep it as test dependency, so we can control the included `Mockito` version. Can you revert `build.gradle` and `gradle.properties`?

raincole:
I don't know, shouldn't we just use whatever version that `mockito-kotlin` is currently using? I'm okay with it, but I wonder why it's important to control `mockito` version ourselves.(For my future reference)

czyzby:
Using a newer minor version is usually OK, as there should be no API changes - there might be some bug fixes or performance improvements, so that's definitely good. The latest `kotlin-mockito` seems to be using slightly outdated `Mockito` version. But yeah, this is such a minor dependency that we could go either way.

Thanks for the contribution.

Issue #58, opened 2017-04-21
Collection operations that return Gdx collections directly


raincole:
Let's say I'm `map`ping a `GdxArray`:

```
val people = gdxArrayOf(Person("Alice"), Person("Bob"))
val names = people.map { it.name }
```

`names` will be a kotlin `List` because `Iterable#map` returns `List`. To get a `GdxArray` I have to call `toGdxArray`. But isn't the whole reason I use `GdxArray` to avoid object creation? 😧 

There seems to be no way to make Kotlin smart enough to build a GdxArray automatically(something like Scala's `canBuildFrom`). Therefore, I'd suggest extension methods like `Iterable#mapGdx`, `Iterable#filterGdx` to make things easier.

I'm willing to write these code and make a PR,  but before I jump into my editor I want to discuss whether it's a good idea.

czyzby:
I'm aware of this issue - completely copying the _stdlib_ API seemed like an overkill for most cases. I don't mind reviewing the code and including this in `gdx-collections` if you're willing to write it, but just know that it can be really time-consuming to cover all operations.

Anyway, `Gdx` suffix is not necessary. For example `GdxArray.map` would override `Iterable.map` method, as long as you keep the `ktx.collections.*` import around. We're already doing this for operator overloading - we'll just add a warning in the `README` file.

czyzby:
It's a shame that LibGDX collections share no common interface. If you want to write such extensions for _all_ collections,  there will be a lot of code duplication.

Common and basic LibGDX collections:
- `Array`
- `ObjectSet`
- `ObjectMap`

Useful, but rarely used collections:
- `IdentityMap`
- `OrderedMap`
- `OrderedSet`
- `SnapshotArray`

Optimized primitives collections:
- `BooleanArray`
- `ByteArray`
- `CharArray`
- `FloatArray`
- `IntArray`
- `IntFloatMap`
- `IntIntMap`
- `IntMap`
- `IntSet`
- `LongArray`
- `LongMap`
- `ObjectFloatMap`
- `ObjectIntMap`

Others:
- `ArrayMap`
- `BinaryHeap`
- `Bits`
- `DelayedRemovalArray`
- `PooledLinkedList` (replaced with KTX `PooledList`)
- `Queue`
- `SortedIntList`

raincole:
I don't plan to rewrite every method tho. Even only `map`, `filter` and probably `flatten` can cover most use cases.

[...]

PR #74, opened 2017-05-21
Collection operations for gdx collections

raincole:
#58 

Add

* `map`
* `filter`
* `flatten`
* `flatMap`

methods to 

* `GdxArray`
* `GdxList`
* `GdxMap`
* `GdxSet`

By the way I feel using `assertEqual` to check elements in a collection is pretty tedious and there must be a way to wrap it as an one-liner like `assertThat(result, SomeMagicalMatcher(listOf(1,2,3)))` or something. But I don't want to do refactor here.

### DIFF
--- a/collections/src/test/kotlin/ktx/collections/arraysTest.kt
+++ b/collections/src/test/kotlin/ktx/collections/arraysTest.kt
@@ -359,9 +359,7 @@ class ArraysTest {
     val result = array.map { it * 2 }
 
     assertTrue(result is GdxArray)
-    assertEquals(2, result[0])
-    assertEquals(4, result[1])
-    assertEquals(6, result[2])
+    assertEquals(GdxArray.with(2, 4, 6), result)
   }
 
   @Test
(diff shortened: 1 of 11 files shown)


collections/src/test/kotlin/ktx/collections/arraysTest.kt
```diff
@@ -352,6 +354,50 @@ class ArraysTest {
...
+    assertEquals(3, result.size)
+    assertEquals(1, result[0])
+    assertEquals(2, result[1])
+    assertEquals(3, result[2])
+  }
+
+  @Test
+  fun `should map elements to lists and flatten them into a new GdxArray`() {
+    val array = GdxArray.with(1, 2, 3)
+    val result = array.flatMap { List(it) { "" }  }
```
czyzby: This is not the best test of `flatMap`, as this does not ensure that each element was transformed. You can use `it` inside `List(it)` body and check if the result is equal to `1, 2, 2, 3, 3, 3`.

collections/src/test/kotlin/ktx/collections/mapsTest.kt
```diff
@@ -378,4 +379,49 @@ class MapsTest {
...
+    assertEquals(3, result.size)
+    assertTrue(1 in result)
+    assertTrue(2 in result)
+    assertTrue(3 in result)
+  }
+
+  @Test
+  fun `should map elements to lists and flatten them into a new GdxArray`() {
+    val map = gdxMapOf("One" to 1, "Two" to 2, "Three" to 3)
+    val result = map.flatMap { List(it.value) { "" }  }
```
czyzby: Same issue as with `GdxArray` test.

PR #85, opened 2017-06-14
DSL extensions for all Engine types

Jkly:
The latest version of Ashley 1.7.3 provides `createEntity()` and `createComponent()` methods for the root level `Engine` implementation. 

This means the DSL can be applied for `Engine` and `PooledEngine` without needing to be `PooledEngine` specific.

One downside is that users of Ashley 1.7.2 and downwards would be incompatible, but 1.7.3 should be a non-breaking upgrade for them.


ashley/src/main/kotlin/ktx/ashley/engines.kt
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Get or create a [Component] by calling [Engine.createComponent].
+ *
+ * @param T the type of [Component] to get or create.
+ * @return an [Component] instance of the selected type.
+ */
+inline fun <reified T : Component> Engine.create(): T = createComponent(T::class.java)
+
+
```
czyzby: Remove one empty line.

czyzby:
I think we should aim at compatibility with the latest dependencies versions, so I'm completely OK with the changes. Can you add a note to the `CHANGELOG.md` file describing the API change?

Jkly:
I've updated the changelog and fixed to formatting issue.

czyzby:
Thanks.

PR #96, opened 2017-08-04
Added raycast extension methods

Jkly:
The ordering of the params in LibGDX `World.rayCast` mean that we're unable to define callbacks using the Kotlin shorthand https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/physics/box2d/World.html#rayCast-com.badlogic.gdx.physics.box2d.RayCastCallback-com.badlogic.gdx.math.Vector2-com.badlogic.gdx.math.Vector2-

These extension methods simply wrap the `rayCast` and puts the callback at the end of the parameter list.


box2d/README.md
```diff
@@ -271,6 +272,19 @@ val body = world.body {
 }
 ```
 
+Creating ray-casts:
+
+```Kotlin
+import ktx.box2d.*
+
+fun createRayCast() {
+  world.rayCast(0f, 0f, 1f, 1f) { fixture, point, normal, fraction ->
```
czyzby: Could you include named parameters in the `rayCast` invocation to make it a bit clearer? 4 subsequent floats are not something I'd consider an obvious API. ; )

box2d/src/main/kotlin/ktx/box2d/worlds.kt
```diff
@@ -60,3 +61,69 @@ fun World.create(bodyDefinition: BodyDefinition): Body {
...
+ *
+ * The lambda should return:
+ * - `-1f`, ignore this fixture and continue
+ * - `0f`, terminate the ray cast
+ * - a fraction, clip the length of the ray to this point
+ * - `1f`, don't clip the ray and continue
+ *
+ * Can be used in place of [com.badlogic.gdx.physics.box2d.RayCastCallback] via Kotlin SAM conversion.
+ */
+typealias RayCastCallback = (Fixture, Vector2, Vector2, Float) -> Float
```
czyzby: While we're at it, the API can be improved by returning `RayCastAction` enum instead of a `Float`. This is something I really hated about the Box2D API. You wouldn't need to declare global constants then. Instead of simply passing the `RayCastCallback` instance to the `World`, you could wrap the callback in our custom adapter that converts the enum result into a `Float` to satisfy Box2D API needs. Remember to `inline` the `rayCast` methods if you decide to go with the enum and the wrappers.

box2d/src/main/kotlin/ktx/box2d/worlds.kt
```diff
@@ -60,3 +61,69 @@ fun World.create(bodyDefinition: BodyDefinition): Body {
...
+ *
+ * @param startX the ray starting point X
+ * @param startY the ray starting point Y
+ * @param endX the ray ending point X
+ * @param endY the ray ending point Y
+ * @param callback a user implemented callback. This is called for every fixture hit.
+ */
+fun World.rayCast(startX: Float, startY: Float, endX: Float, endY: Float, callback: RayCastCallback) {
+  rayCast(callback, startX, startY, endX, endY)
+}
```
czyzby: Missing new line at the end of the file.

box2d/src/test/kotlin/ktx/box2d/worldsTest.kt
```diff
@@ -171,4 +171,44 @@ class WorldsTest : Box2DTest() {
     assertSame(body.fixtureList[0], callbackParameter)
     world.dispose()
   }
+
+  @Test
+  fun `should ray-cast between two vectors`() {
+    val world = createWorld()
+
+    val expectedEdge = world.body {}.edge(from = Vector2.Zero, to = Vector2(0f, 2f)) {}
```
czyzby: I think readability of this declaration can be improved with:

```Kotlin
    world.body {
      edge(from = Vector2.Zero, to = Vector2(0f, 2f)) {}
    }
```

Don't worry too much about this not being a one-liner. Same thing about the other test.

box2d/src/main/kotlin/ktx/box2d/worlds.kt
czyzby: Note that this alias clashes with `RayCastCallback` interface from Box2D. It might be a good idea to choose a different name.

Jkly: I agree the global constants aren't ideal; the problem is that the returned Float has a special meaning in Box2D if it's a fraction.

For example, to get only the nearest raycast hit in the most efficient way, you have to return the fraction and then pick the nearest intersection from the multiple calls to the callback (since there is not guarantee that the hits are returned in order of nearest to furthest).

It's best explained here! http://www.iforce2d.net/b2dtut/world-querying

box2d/src/test/kotlin/ktx/box2d/worldsTest.kt
Jkly: I chose to call `edge()` on the returned body because the test needs a reference to the created Edge fixture.

I couldn't see an alternative, unless I make `var expectedEdge` or a callback, but both those options don't seem ideal.

box2d/README.md
Jkly: yep, great point 👍

box2d/src/main/kotlin/ktx/box2d/worlds.kt
czyzby: I see. Can you at least aggregate the constants into an `object` properties, so they are not available in the global scope? I'd rather use `RayCastResult.CONTINUE` (for example) than a global constant.

Jkly: Yep - that sounds like a good idea 👍

Jkly:
@czyzby Thanks for the comments - I think I've addressed them all, could you take another look?

czyzby:
Thanks. The new snapshot will be released soon.

Issue #94, opened 2017-06-27
Default functional parameters in inlined builder methods


czyzby:
- [x] Scene2D builders.
- [x] Scene2D style builders (`ktx-style`, `ktx-vis-style`).
- [x] Box2D DSL.
- [x] Ashley DSL.

--- 

[Kotlin 1.2](https://blog.jetbrains.com/kotlin/2017/06/early-access-program-for-kotlin-1-2-has-been-started/) brings inline functions with optional functional parameters. Currently `ktx-scene2d` defines two factory methods for each widget: one with and one without the building block, as you could not simply make the building block optional with `init: Actor.() -> Unit = {}` without removing the `inline` keyword. When Kotlin 1.2 comes out, we can refactor the existing actors' factory methods to use default no-op building blocks and remove the no-init-block methods - without breaking the existing KTX applications.

```Kotlin
// Kotlin 1.1:
inline fun KWidget<*>.button(
    style: String = defaultStyle,
    skin: Skin = Scene2DSkin.defaultSkin) = appendActor(KButton(skin, style))

inline fun <S> KWidget<S>.button(
    style: String = defaultStyle,
    skin: Skin = Scene2DSkin.defaultSkin,
    init: KButton.(S) -> Unit) = actor(KButton(skin, style), init)

// Kotlin 1.2:
inline fun <S> KWidget<S>.button(
    style: String = defaultStyle,
    skin: Skin = Scene2DSkin.defaultSkin,
    init: KButton.(S) -> Unit = {}) = actor(KButton(skin, style), init)
```

czyzby:
Note that this might also be useful for Box2D body/fixture builders.

czyzby:
@Jkly `ktx-ashley` could also benefit from this feature, if I'm not mistaken. Can you take a look?

czyzby:
Inlining text tooltip factory methods [seems to cause compiler errors](https://github.com/libktx/ktx/commit/2f54ebafe31a5eb332c8452487b02510f3e19c64#diff-73905c72eb23c7cba79c23c4d1d41425R6).

```java
org.jetbrains.kotlin.codegen.CompilationException: Back-end (JVM) Internal error: Couldn't inline method call 'addTooltip' into
@org.junit.Test public final fun `should add Tooltip with init block`(): kotlin.Unit defined in ktx.scene2d.TooltipFactoriesTest
@Test
  fun `should add Tooltip with init block`() {
    val actor = Actor()
[...]

PR #115, opened 2017-11-29
Use default functional parameters for ktx-ashley

Jkly:
https://github.com/libktx/ktx/issues/94


czyzby:
This causes no API changes, right? It seems you haven't used the KTX code formatter, but it's OK - I'll reformat the file and update CHANGELOG. Thanks.

Jkly:
Yep the API should stay the same.

PR #122, opened 2018-01-25
Add higher-order function parameters to bindSingleton() functions.

TomGrill:
I recommend to swap the bindSingleton parameters and put the "vararg to: Class<out Type>" as first paramter, like it is already done for the bind() functions.

Also added higher-order function parameters which allow now to use lambda expression for bindSingleton() as well.

F.e.

```
bindSingleton {
     if (something) AndroidImplementation() else DesktopImplementation()
}

bindSingleton (Implementation::class.java) {
     if (something) AndroidImplementation() else DesktopImplementation()
}
```

### DIFF
--- a/inject/src/main/kotlin/ktx/inject/inject.kt
+++ b/inject/src/main/kotlin/ktx/inject/inject.kt
@@ -141,6 +141,13 @@ open class Context : Disposable {
   inline fun <reified Type : Any> bindSingleton(singleton: Type) = bind(SingletonProvider(singleton))
 
   /**
+   * Allows to bind a singleton to the chosen class.
+   * @param provider will be invoked and the return value converted to a provider that always returns the same instance.
+   * @throws InjectionException if provider for the selected type is already defined.
+   */
+  inline fun <reified Type : Any> bindSingleton(provider: () -> Type) = bind(SingletonProvider(provider()))
+
+  /**
    * Allows to bind a provider to multiple classes in hierarchy of the provided instances class.
    * @param to list of interfaces and classes in the class hierarchy of the objects provided by the provider. Any time
    *    any of the passed classes will be requested for injection, the selected provider will be invoked.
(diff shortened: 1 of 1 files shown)


inject/src/main/kotlin/ktx/inject/inject.kt
```diff
@@ -151,13 +158,23 @@ open class Context : Disposable {
...
+   * Allows to bind a provider instance to multiple classes in its hierarchy.
+   * @param to list of interfaces and classes in the class hierarchy of the provider. Any time any of the passed classes
+   *    will be requested for injection, the selected provider will be returned.
+   * @param provider provides instances of classes compatible with the passed types.
    * @throws InjectionException if provider for any of the selected types is already defined.
    */
-  fun <Type : Any> bindSingleton(singleton: Type, vararg to: Class<out Type>)
-      = bind(*to, provider = SingletonProvider(singleton))
+  fun <Type : Any> bindSingleton(vararg to: Class<out Type>, provider: () -> Type)
+          = bind(*to, provider = SingletonProvider(provider()))
```
czyzby: Use KTX formatter. We will migrate to official Kotlin formatting style, but I'd like us to stay consistent for now.

inject/src/main/kotlin/ktx/inject/inject.kt
```diff
@@ -151,13 +158,23 @@ open class Context : Disposable {
...
+  /**
+   * Allows to bind a provider instance to multiple classes in its hierarchy.
+   * @param to list of interfaces and classes in the class hierarchy of the provider. Any time any of the passed classes
+   *    will be requested for injection, the selected provider will be returned.
+   * @param provider provides instances of classes compatible with the passed types.
    * @throws InjectionException if provider for any of the selected types is already defined.
    */
-  fun <Type : Any> bindSingleton(singleton: Type, vararg to: Class<out Type>)
-      = bind(*to, provider = SingletonProvider(singleton))
+  fun <Type : Any> bindSingleton(vararg to: Class<out Type>, provider: () -> Type)
```
czyzby: Since you immediately invoke the provided, please inline this method.

inject/src/main/kotlin/ktx/inject/inject.kt
```diff
@@ -151,13 +158,23 @@ open class Context : Disposable {
...
    * @param to list of interfaces and classes in the class hierarchy of the singleton. Any time any of the passed classes
    *    will be requested for injection, the selected singleton will be returned.
+   * @param singleton instance of class compatible with the passed types.
+   * @throws InjectionException if provider for any of the selected types is already defined.
+   */
+  fun <Type : Any> bindSingleton(vararg to: Class<out Type>, singleton: Type)
+          = bind(*to, provider = SingletonProvider(singleton))
+
+  /**
+   * Allows to bind a provider instance to multiple classes in its hierarchy.
```
czyzby: It should be more like `Allows to bind the result of the provider to multiple classes in its hierarchy.`

TomGrill:
Alright, i will do a new PR to the develop branch.

PR #130, opened 2018-02-24
develop-math

mrdlink:
Math library didn't work as expected. Binary operators modified the objects given in as parameters. Therefore new assign operators are introduced, and binary operators are fixed the way they are expected to work.
Swizzling properties and alternative coordinate names are introduced to the Vector2 and Vector3, just like GLSL, for better syntax.
Matrix vector multiplication operators are altered. Now, for example matrix4 * vector3 returns new Vector3.

Codebreaking changes may be applied since extension functions are altered in their behaviour and/or removed completely if they don't make sense in a mathematic way.


czyzby:
You introduced two changes in the API in a single pull request - while refactoring of the operators (with the `Assign` methods introduction) is something I could agree with, I'm not sure about the extension properties to the vector classes.

I see where you're coming from, but I feel like `Vector2`/`Vector3` are more general purpose in LibGDX and rarely used directly for graphics programming - at least not by most of the users. Single-character properties might clutter the API and confuse new developers. This feels like a very problem-specific extension that not everyone would use or even want in their application. Correct me if I'm wrong.

Anyway, it would be great if you could describe the changes in the [changelog](https://github.com/libktx/ktx/blob/develop/CHANGELOG.md) under `1.9.8-SNAPSHOT` and remove the vector extension properties for now. I can merge this as is and apply the changes myself if you don't have the time.

czyzby:
Thanks, I'll upload snapshot release with your changes soon.

czyzby:
I added missing methods to the changelog. I also removed the `Color` factory method, as its not really connected with math - such extension could be added to a separate module like `ktx-graphics`. Thanks again, the API is more sensible after these changes.

PR #131, opened 2018-03-07
ktx-box2d: extension method for querying

Jkly:
This extension method allows querying for fixtures overlapping an AABB in a more Kotlin idiomatic way.


box2d/src/main/kotlin/ktx/box2d/worlds.kt
```diff
@@ -152,3 +152,36 @@ fun World.rayCast(
...
+ * Query the world for all fixtures that potentially overlap the provided AABB (Axis-Aligned Bounding Box).
+ *
+ * @param lowerX the x coordinate of the lower left corner
+ * @param lowerY the y coordinate of the lower left corner
+ * @param upperX the x coordinate of the upper right corner
+ * @param upperY the y coordinate of the upper right corner
+ * @param callback a user implemented callback that is called for every fixture overlapping the AABB.
+ * @see Query
+ */
+fun World.query(lowerX: Float, lowerY: Float, upperX: Float, upperY: Float, callback: KtyQueryCallback) {
```
czyzby: Typo in `KtyQueryCallback`.

box2d/src/main/kotlin/ktx/box2d/worlds.kt
```diff
@@ -152,3 +152,36 @@ fun World.rayCast(
...
+   */
+  const val STOP = false
+  /**
+   * Continue querying for the next match.
+   * @see KtyQueryCallback
+   */
+  const val CONTINUE = true
+}
+
+typealias KtyQueryCallback = (Fixture) -> Boolean
```
czyzby: Rename and document the alias.

box2d/src/test/kotlin/ktx/box2d/worldsTest.kt
```diff
@@ -218,4 +218,39 @@ class WorldsTest : Box2DTest() {
     assertEquals(0f, RayCast.TERMINATE, tolerance)
     assertEquals(-1f, RayCast.IGNORE, tolerance)
   }
+
+  @Test
+  fun `should query for overlapping fixtures with AABB and stop`() {
+    val world = createWorld()
+    val matchingEdge1 = world.body {}.edge(from = Vector2.Zero, to = Vector2(0f, 2f)) {}
+    val matchingEdge2 = world.body {}.edge(from = Vector2(1f, 0f), to = Vector2(1f, 2f)) {}
```
czyzby: Is it still necessary to use empty `{}` instead of just `()`? I'm pretty sure KTX Box2D provides default init blocks now - if this is impossible, this is a regression or omission in the API.

box2d/src/test/kotlin/ktx/box2d/worldsTest.kt
```diff
@@ -218,4 +218,39 @@ class WorldsTest : Box2DTest() {
...
+    val matchedFixtures = mutableSetOf<Fixture>()
+
+    world.query(-1f, 1f, 1f, 1f) { fixture ->
```
czyzby: Use named parameters. Tests showcase the API, 4 unnamed floats can be somewhat confusing.

box2d/src/test/kotlin/ktx/box2d/worldsTest.kt
```diff
@@ -218,4 +218,39 @@ class WorldsTest : Box2DTest() {
...
+      matchedFixtures += fixture
+      Query.CONTINUE
+    }
+
+    assertEquals(2, matchedFixtures.size)
+    assertTrue(matchedFixtures.contains(matchingEdge1))
+    assertTrue(matchedFixtures.contains(matchingEdge2))
```
czyzby: You can just perform `assertEquals` with two sets instead of comparing the size and 2 `contains` checks.

box2d/src/test/kotlin/ktx/box2d/worldsTest.kt
```diff
@@ -218,4 +218,39 @@ class WorldsTest : Box2DTest() {
...
+  }
+
+  @Test
+  fun `should query for overlapping fixtures with AABB and continue`() {
+    val world = createWorld()
+    val matchingEdge1 = world.body {}.edge(from = Vector2.Zero, to = Vector2(0f, 2f)) {}
+    val matchingEdge2 = world.body {}.edge(from = Vector2(1f, 0f), to = Vector2(1f, 2f)) {}
+    val matchedFixtures = mutableSetOf<Fixture>()
+
+    world.query(-1f, 1f, 1f, 1f) { fixture ->
```
czyzby: Use named parameters.

box2d/src/test/kotlin/ktx/box2d/worldsTest.kt
```diff
@@ -218,4 +218,39 @@ class WorldsTest : Box2DTest() {
...
+    assertEquals(1, matchedFixtures.size)
+    assertTrue(matchedFixtures.contains(matchingEdge1) || matchedFixtures.contains(matchingEdge2))
+    world.dispose()
+  }
+
+  @Test
+  fun `should query for overlapping fixtures with AABB and continue`() {
+    val world = createWorld()
+    val matchingEdge1 = world.body {}.edge(from = Vector2.Zero, to = Vector2(0f, 2f)) {}
+    val matchingEdge2 = world.body {}.edge(from = Vector2(1f, 0f), to = Vector2(1f, 2f)) {}
```
czyzby: See if you can use simple method call `()` instead of an empty init block `{}`.

box2d/src/test/kotlin/ktx/box2d/worldsTest.kt
```diff
@@ -218,4 +218,39 @@ class WorldsTest : Box2DTest() {
...
+    world.query(-1f, 1f, 1f, 1f) { fixture ->
+      matchedFixtures += fixture
+      Query.CONTINUE
+    }
+
+    assertEquals(2, matchedFixtures.size)
+    assertTrue(matchedFixtures.contains(matchingEdge1))
+    assertTrue(matchedFixtures.contains(matchingEdge2))
+    world.dispose()
+  }
```
czyzby: Can you add another test for a query that returns no fixtures to make sure that it does not break on this "edge case"? (E.g. returns null instead of empty collection.)

box2d/src/test/kotlin/ktx/box2d/worldsTest.kt
Jkly: Looks like they defaults are missing for `World.body`

czyzby: Good catch, I'll add it today.

czyzby:
Thanks.

PR #158, opened 2018-09-20
Throw an appropriate exception if createComponent fails

Jkly:
Note that PooledEngine already throws a RuntimeException with an appropriate message "Class cannot be created (missing no-arg constructor)", instead of returning null.

`Engine.create` still catches the Exception and throws a`CreateComponentException` for consistency between the two Engine types.


czyzby:
Other than some cosmetic issues with long exception class declaration, it's great. Thanks for some tests to cover this use case.

czyzby:
I've noticed you made the PR into the `master` branch rather than `develop`. I won't revert the changes, but please add new features to `develop` first.

Jkly:
Ah yes, sorry about that @czyzby!

czyzby:
It's OK, thanks again for the quick fix. ;)

Issue #152, opened 2018-08-13
Immutable vector / colors


jcornaz:
Hi, 

Thank you very much for providing theses utilities for Kotlin :-)

I would like to propose to introduce some immutable classes for at least vector and colors. (could be called `KVector2` or `ImmutableVector2` for instance)

In order to allow easy usage with LibGDX they could either:
* inherit from their equivalent in LibG, deprecate all mutation method and throw if a mutation method is called
* provide adapter like `ImmutableVector2.asVector2()` and `Vector2.toImmutableVector2()`

**Why?**
Even in Java in find error-prone that classes like vector or color are mutable. It is for instance possible to change by mistake `Vector2.ZERO`, and it is a mistake really easy to do. But I think that's even more true in Kotlin where the language provide good support for immutability (`val` is the default, `data class` provides `copy` to easily get a new object, immutable fields can be smart-casted. 

In other words:
* Even in Java it should have been immutabl
* In Kotlin, immutable data classes are even more idiomatic.

czyzby:
I think LibGDX sacrifices safety for conciseness and performance - the difference might not be noticeable on desktop, but it does somewhat matter on mobile.

That being said, immutable vectors and colors would be a great addition and might prevent some non-obvious bugs - but since there is no `Vector` or `Color` interface that you could just implement, the conversion code interacting with existing LibGDX APIs could quickly become tedious. We might also have to reimplement all (or most) methods, or at least delegate the to the original, mutable objects.

@jcornaz Would you like to try implementing such immutable classes or is it a feature request?

jcornaz:
>I think LibGDX sacrifices safety for conciseness and performance - the difference might not be noticeable on desktop, but it does somewhat matter on mobile.

Yes, it does. However as you said it is less true on desktop. And I think it may even actually lead to worse performances, as it is dangerous to do share mutable vector data-structure among different threads.

> Would you like to try implementing such immutable classes or is it a feature request?

This is a feature request. But I don't mind trying to implement it myself and propose a PR.

czyzby:
I'd be happy to review the PR, thanks.

jcornaz:
What would you think about the following design:

```kotlin
import com.badlogic.gdx.math.Vector2
import kotlin.math.sqrt

data class ImVector2(val x: Float = 0f, val y: Float = 0f) {
  val isZero get() = x == 0f && y == 0f

  // immutability allow some performance improvements as well, by making able to cache results:
[...]

PR #161, opened 2018-09-30
Immutable Vector 2D

jcornaz:
Contributes to #152

I open this pull request, because I consider `ImmutableVector2` to be feature-complete.

### Before merge I want to:
* [x] Rename for `ImmutableVectorN`
* [x] Inline *alias-functions*
* [x] Add tiny vectors (like `(0.001, -0.001)`) in the list of vector to test
* [x] Add the missing comments
* [x] Replicate syntax sugar KTX was providing for `Vector2`
* [x] Update the changelog
* [x] Update math module's readme
* [x] Explicitly test equals and hashCode methods. (It is very important, since most of the tests rely on it)
* [x] Re-read everything / Fix typos
* [x] Add examples to the README section
* [x] Add some test against expected values in order to make sure to detect bugs present in both KTX and LibGDX
* [x] Fix documentation and address requested changes

### Open questions and points to notice :
* ~~What name should we choose? I know I proposed it, but I don't like `KVector`, because the `K` means nothing, and certainly not "immutable". I think I'd prefer `ImVector` or `ImmutableVector`~~
  * Solution: We'll name them `ImmutableVectorN` and write in the doc that one can create typealiases for shorter names
* ~~Should the 3D vectors be part of this PR?~~
  * Solution: Vector 3D will be part of a separate PR
* ~~I intentionnaly chose to make `angle()` return values between `-180` and `180` instead of between `0` and `360` like LibDGX does. I chose that to be consistent with `angle(Vector2)`, `angleRad()` and `angleRad(Vector2)`.~~
  * Update: We provide `angleDeg` functions and deprecate `angle` to make the difference in behavior clearer and less error-prone
* Two tests are disabled while waiting the resolution of https://github.com/libgdx/libgdx/issues/5385
* Most of the function got exact same name as their equivalent in `Vector2`. Exceptions to this rule are setter and functions which became Kotlin operators:
  * `setLength` and `setLength2` became `withLength` and `withLegnth2`
  * `setToRandomRotation` became `withRandomRotation`
  * `scl` became the operator `times`: `v * 3` or `v1 * v2`
    * `unaryMinus` is the equivalent of `scl(-1f)`: `-v == v * -1`
  * `add` became the operator `plus`: `v1 + v2`
  * `sub` became the operator `minus`: `v1 - v2`


czyzby:
Looking good so far, thank you for working on the tests right away.

KVector does sound non-obvious, but it is Kotlin's convention to prefer immutability and prefix mutable variants with `Mutable` rather than the other way around. I think it sounds reasonable to make an exception here and name them `ImmutableVectorN`, with a note in the documentation on how to quickly make an alias - possibly with a few examples:

```kotlin
typealias Vec2 = ImmutableVector2
typealias KVector2 = ImmutableVector2
```

> Should the 3D vectors be part of this PR?

If you consider 2D vector implementation feature complete, I don't see why we wouldn't release a snapshot release with what we've got so far.

math/src/main/kotlin/ktx/math/KVector.kt
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+fun <T : KVector<T>> T.withLength(len: Float): T = withLength2(len * len)
+
+/** @return The opposite vector of same length */
+operator fun <T : KVector<T>> T.unaryMinus(): T = times(-1f)
+
+fun <T : KVector<T>> T.interpolate(target: T, alpha: Float, interpolator: Interpolation): T =
+        lerp(target, interpolator.apply(alpha))
```
jcornaz: Should I inline this kind of function?

I know many people likes inlining functions which are not more than aliases. But the gain in performance is usually between tiny and not measurable, and inline functions make implementation changes not possible without breaking binary compatibility.

jcornaz:
>variants with Mutable rather than the other way around

Yes. However when not possible, even the Kotlin team prefix with `Immutable`: https://github.com/Kotlin/kotlinx.collections.immutable/tree/master/kotlinx-collections-immutable/src/main/kotlin/kotlinx/collections/immutable

Use proper explicit name and propose the user to create typealiases is a great solution, I love it! Thanks.

math/src/main/kotlin/ktx/math/KVector.kt
czyzby: On desktop JVMs, the differences might not be noticeable, but on Android inlined functions lower the total method count and might slightly improve the performance if the function is called very often. As a rule of thumb, functions with lambda params are usually inlined in KTX.

jcornaz: I haven't thought about method count indeed.

> As a rule of thumb, functions with lambda params are usually inlined in KTX

There's no lambda involved in this case. But they are *aliases*. For instance `-v` is an alias for `v * -1`. So the function could be safely inlined (reducing method count) without worrying to much about binary compatibility, or debugability (since inline function don't exist at runtime).

The argument of method counts makes me think it'd be better to inline alias-functions.

czyzby: After a quick look on the code you commented, I originally thought that interpolator is a function object and would be inlined. As I've said, I prefer that all KTX methods with lambda parameters  (and reified generics, of course) are inlined, but in other cases feel free to mark methods and properties as you consider sensible.

jcornaz:
I have a concern about the `angle()` function which between `[-180, +180]` instead of `[0, 360]` of LibGDX.

I made this choice in order to make it consistent with `angle(Vector2)`, `angleRad()` and `angleRad(Vector2)`.

But I wonder if I shouldn't rename them `angleDeg` in order to make it clear that they aren't equivalent to the LibGDX ones.

czyzby:
I prefer more explicit names either way, so I don't see a problem with `angleDeg`. Do you want to add a deprecated `angle` method that invokes `angleDeg`, so that existing code converted to `ImmutableVector` works out of the box? It could point out that you should use the explicitly named variant to prevent bugs.

jcornaz:
Good idea to add a deprecated `angle` function working out of the box, providing explicit explaination and quick-fix from intelliJ. Great thanks again.

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 #### 1.9.8-SNAPSHOT
+- **[FEATURE]** Added `ImmutableVector2`, an immutable equivalent to `Vector2`.
```
czyzby: Add the module name as well so that it's obvious where the class was added; see the previous feature logs.

jcornaz:
I think I'm done @czyzby.

Please let me know if you find any potential issue, if you want me to add more test, or if you have any question about the code I wrote.

czyzby:
Thanks, I'll try to do a full review tomorrow.

math/README.md
```diff
@@ -39,6 +39,28 @@ Note that since `Shape2D` has `contains(Vector2)` method, `in` operator can be u
...
+- `Vector2.toImmutableVector2()` Returns an immutable vector with same `x` and `y` attributes than this `Vector2`
+- `ImmutableVector2.toVector2()` Returns an mutable vector with same `x` and `y` attributes than this `ImmutableVector2`
+- Notable difference with `Vector2`
+  - `+`, `-`, `*`, `/` are available and replace `add`, `sub` and `scl`.
+  - `withLength()` and `withLength2()` replace `setLength()` and `setLength2()` and return a new vector of same direction
+  with the specified length
+  - `withRandomRotation` replace `setToRandomRotation` and return a new vector of same length and a random rotation
+  - `withAngleDeg()` and `withAngleRad` replace `setAngle` and `setAngleRad` and return a new vector of same length and
+  the given angle to x-axis
+  - `cpy` is not provided and is not necessary. Immutable vectors can be safely shared.
```
czyzby: Add the deprecated `cpy` method returning `this` for compatibility with the original `Vector2`. `ImmutableVector2` should be a (mostly) drop-in replacement.

math/README.md
```diff
@@ -39,6 +39,28 @@ Note that since `Shape2D` has `contains(Vector2)` method, `in` operator can be u
...
 call `vec in rect` (or `vec !in rect`) to check if the rectangle contains (or doesn't) the point stored by the vector.
 
+#### `ImmutableVector2`
+- `ImmutableVector2` is an immutable equivalent to `Vector2`. It provides most of the functionality of `Vector2`, but
+mutation methods return new vectors instead of mutate the reference.
+- Note that one may want to create type aliases to makes the usage more concise: `typealias Vect2 = ImmutableVector2`
+- `ImmutableVector` is comparable (`>`, `>=`, `<`, `<=` are available). Comparison is evaluated by length
+- instances can be destructed: `val (x, y) = vector2`
+- `Vector2.toImmutableVector2()` Returns an immutable vector with same `x` and `y` attributes than this `Vector2`
+- `ImmutableVector2.toVector2()` Returns an mutable vector with same `x` and `y` attributes than this `ImmutableVector2`
```
czyzby: I think `toImmutable` and `toMutable` might be better names for conversion methods.

math/README.md
```diff
@@ -39,6 +39,28 @@ Note that since `Shape2D` has `contains(Vector2)` method, `in` operator can be u
...
+- Notable difference with `Vector2`
+  - `+`, `-`, `*`, `/` are available and replace `add`, `sub` and `scl`.
+  - `withLength()` and `withLength2()` replace `setLength()` and `setLength2()` and return a new vector of same direction
+  with the specified length
+  - `withRandomRotation` replace `setToRandomRotation` and return a new vector of same length and a random rotation
+  - `withAngleDeg()` and `withAngleRad` replace `setAngle` and `setAngleRad` and return a new vector of same length and
+  the given angle to x-axis
+  - `cpy` is not provided and is not necessary. Immutable vectors can be safely shared.
+  - `set(x, y)` and `setZero()` are not provided.
```
czyzby: You can add a `with` method (to replace `set`) that provides default named params equal to current vector state. Sample usage:

```kotlin
vector.with(x = 4f)
```

EDIT: I noticed we can also use `copy` from Kotlin's data class definition instead, so you can just point that out in the readme.

math/src/main/kotlin/ktx/math/ImmutableVector.kt
```diff
@@ -0,0 +1,154 @@
...
+    /** Compares this vector with the other vector, using the supplied epsilon for fuzzy equality testing */
+    fun epsilonEquals(other: T, epsilon: Float): Boolean
+
+    override fun compareTo(other: T): Int = len2.compareTo(other.len2)
+}
+
+/**
+ * Returns the euclidean length
+ */
+inline val ImmutableVector<*>.len: Float get() = sqrt(len2)
```
czyzby: Why is this property inlined?

math/src/main/kotlin/ktx/math/ImmutableVector.kt
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Returns the unit vector of same direction or this vector if it is zero.
+ */
+val <T : ImmutableVector<T>> T.nor: T
+    get() {
+        val l2 = len2
+
+        if (l2 == 1f || l2 == 0f) return this
```
czyzby: Convert to `if`+`else` expression, return it right away. This isn't very idiomatic. Same goes for the method below.

Why is this property outside of the interface?

math/src/main/kotlin/ktx/math/ImmutableVector.kt
```diff
@@ -0,0 +1,154 @@
...
+inline fun <T : ImmutableVector<T>> T.isUnit(margin: Float = MathUtils.FLOAT_ROUNDING_ERROR): Boolean = abs(1f - len2) < margin
+
+/** Returns the distance between this and the [other] vector */
+inline infix fun <T : ImmutableVector<T>> T.dst(other: T): Float = sqrt(dst2(other))
+
+/** Returns this vector scaled by (1 / [scalar]) */
+inline operator fun <T : ImmutableVector<T>> T.div(scalar: Float): T = times(1 / scalar)
+
+/** Returns this vector if the [ImmutableVector.len] is <= [limit] or a vector with the same direction and length [limit] otherwise */
+inline fun <T : ImmutableVector<T>> T.limit(limit: Float): T = limit2(limit * limit)
```
czyzby: Not sure if most of these methods should be inlined and outside of the interface. As I've mentioned, we usually inline functions with lambda parameters or reified generics. Is there a reason you inlined most of this outside of reducing method count?

math/src/main/kotlin/ktx/math/ImmutableVector.kt
```diff
@@ -0,0 +1,154 @@
...
+ * Returns the euclidean length
+ */
+inline val ImmutableVector<*>.len: Float get() = sqrt(len2)
+
+/**
+ * Returns the unit vector of same direction or this vector if it is zero.
+ */
+val <T : ImmutableVector<T>> T.nor: T
+    get() {
+        val l2 = len2
```
czyzby: Can you rename the `l2` parameter to `lengthSquared` for readability? I want the sources to be easy to read for Kotlin/LibGDX beginners.

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,182 @@
...
+    fun plus(deltaX: Float = 0f, deltaY: Float = 0f): ImmutableVector2 = ImmutableVector2(x + deltaX, y + deltaY)
+
+    override fun inc(): ImmutableVector2 = copy(x = x + 1, y = y + 1)
+    override fun dec(): ImmutableVector2 = copy(x = x - 1, y = y - 1)
+
+    override operator fun times(scalar: Float): ImmutableVector2 = times(scalar, scalar)
+    override operator fun times(vector: ImmutableVector2): ImmutableVector2 = times(vector.x, vector.y)
+
+    /** Returns this vector scaled by the given [xf] and [yf] factors */
+    fun times(xf: Float, yf: Float): ImmutableVector2 = ImmutableVector2(x * xf, y * yf)
```
czyzby: I understand that these were the method parameters in the original LibGDX code, but I'm not a fan of such non-obvious names. You could use `x`, `factorX` or `xFactor` instead.

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,182 @@
...
+
+    override fun dot(vector: ImmutableVector2): Float = dot(vector.x, vector.y)
+
+    /** Returns the dot product of this vector by the given vector */
+    fun dot(ox: Float, oy: Float): Float = Vector2.dot(x, y, ox, oy)
```
czyzby: Again, non-obvious names. `x`, `otherX`?

czyzby: Same goes for a few methods below.

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * This method is faster than [dst] because it avoids calculating a square root. It is useful for comparisons,
+     * but not for getting exact distance, as the return value is the square of the actual distance.
+     */
+    fun dst2(ox: Float, oy: Float): Float = Vector2.dst2(x, y, ox, oy)
+
+    /** @return the distance between this and the other vector */
+    fun dst(ox: Float, oy: Float): Float = Vector2.dst(x, y, ox, oy)
+
+    operator fun times(matrix: Matrix3): ImmutableVector2 = ImmutableVector2(
```
czyzby: Missing documentation.

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,182 @@
...
+    override fun toString(): String = "($x,$y)"
+
+    @Deprecated(
+            message = "This function doesn't behave like its equivalent in LibGDX and return an angle between -180 and 180 (some LibGDX functions return between -180 and 180 and some other between 0 and 360)",
+            replaceWith = ReplaceWith("angleDeg(reference)")
+    )
+    inline fun angle(reference: ImmutableVector2 = ImmutableVector2.X): Float = angleDeg(reference)
+
+    @Deprecated(
+            message = "use rotateDeg instead. (this function is not provided to be consistent with angleDeg)",
```
czyzby: > not provided

Not guaranteed?

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,182 @@
...
+        /**
+         * Returns the [ImmutableVector2] represented by the specified string according to the format of [ImmutableVector2::toString]
+         */
+        fun fromString(string: String): ImmutableVector2 =
+                Vector2().fromString(string).toImmutableVector2()
+    }
+}
+
+/** @return an instance of [ImmutableVector2] with the same x and y values */
+inline fun ImmutableVector2.toVector2(): Vector2 = Vector2(x, y)
```
czyzby: I'd name it `toMutable` instead.

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,182 @@
...
+
+/** @return an instance of [Vector2] with the same x and y values */
+inline fun Vector2.toImmutableVector2(): ImmutableVector2 = ImmutableVector2(x, y)
```
czyzby: I'd name it `toImmutable`.

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,182 @@
...
+/** Returns a vector of same length rotated by the given [angle] in degree */
+inline fun ImmutableVector2.rotateDeg(angle: Float): ImmutableVector2 =
+        rotateRad(angle * MathUtils.degreesToRadians)
+
+/**
+ * Returns a vector of same length rotated by 90 degrees in the given [dir]
+ *
+ * @param dir positive value means toward positive y-axis (typically counter-clockwise). Negative value means toward negative y-axis (typically clockwise).
+ */
+inline fun ImmutableVector2.rotate90(dir: Int): ImmutableVector2 =
```
czyzby: `direction`.

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,182 @@
...
+
+/** Returns a vector of same length with the given [angle] in degree */
+inline fun ImmutableVector2.withAngleDeg(angle: Float): ImmutableVector2 =
+        withAngleRad(angle * MathUtils.degreesToRadians)
+
+/** Calculates the 2D cross product between this and the [other] vector */
+inline infix fun ImmutableVector2.x(other: ImmutableVector2): Float = crs(other.x, other.y)
+
+/** Calculates the 2D cross product between this and the [other] vector */
+inline infix fun ImmutableVector2.crs(other: ImmutableVector2): Float = crs(other.x, other.y)
```
czyzby: Again, I'm not sure all of these should be inlined.

czyzby:
Math module didn't use to have an examples section, since it mostly contained syntax sugar, new operator definitions and so on. With the addition of immutable data classes, I'd add a few examples to the README including:

* Initiating a new `ImmutableVector2`.
* Converting `Vector2` to `ImmutableVector2`.
* Adding type alias for `ImmutableVector2` (e.g. `Vec2`).

math/src/main/kotlin/ktx/math/ImmutableVector.kt
jcornaz: Because it is only an alias of `sqrt(len2)`.

math/src/main/kotlin/ktx/math/ImmutableVector.kt
jcornaz: >Convert to if+else expression, return it right away. This isn't very idiomatic. Same goes for the method below.

Ok

>Why is this property outside of the interface?

Because it is only an helper available for any `ImmutableVector`. Writing it outside of the interface bring to advantages:
* No need to rewrite it for Vector3
* Future proof (It is easy and safe to move from extension to member, not the opposite)

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
jcornaz: They are only aliases.

Of course the policy of inlining aliases is debatable and I'll remove the inline for all functions if you want.

I just thought: the less method count the better. So if an function is only an *alias* (like here) I'd inline.

Just tell me if you want me to remove the aliases, and I'll do it.

jcornaz:
Thanks @czyzby. I'll do the requested changes tonight.

About inline modifier, it is mainly up to you. I did what I personally consider best. But this is very opinionated I think, and I'll just remove the `inline` modifier if you ask me to.

Advantages of inline:
* don't increase method count
* don't have indirection

Disadvantages of inline:
* content is not encapsulated (implementation details are exposed and are harder to change in the future)

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
jcornaz: ooops

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
jcornaz: >I'm not a fan of such non-obvious names

Me neither. Actually in LibDGX they were named `x` and `y`. But i tend to find it even less obvious.

I'll go for `factorX`. (same for the method bellow)

math/src/main/kotlin/ktx/math/ImmutableVector.kt
jcornaz: Because I consider them aliases. I don't consider them to have their own proper logic. Therefore the main problem of inline function which is the fact they expose their implementation, does not apply.

But as I said in my other comments, I'll remove the inline modifier, if you prefer.

jcornaz:
I have a small question about deprecated function added to smoother migration from `Vector2`:

Should we also add (deprecated) method which are explicitly setters? For instance `setAngle()` could be there deprecated with a quick fix for `withAngle`.

(Note I'am not advocating for or against this option. I only put it on the table)

Also another point, (somehow related) is that some functions are provided with exact same signature than in `Vector2`, but mutate in the case with `Vector2` and not in the case of `ImmutableVector2`.
This is the case for instance of `lerp`. Should we rename them to avoid confusion?

math/src/main/kotlin/ktx/math/ImmutableVector.kt
czyzby: I see a single reason why you would want these methods to be within the interface - you can provide your own implementation. Most immutable vector methods are deterministic and depend on the immutable state of the vector (e.g. length, angle), so in some cases people might want to cache these values e.g. by introducing a new lazy field. I prefer thin interfaces as well, so I won't try to argue, but consider if any of the properties/methods you introduced outside of the interface or class definitions might need overriding.

czyzby:
> Should we also add (deprecated) method which are explicitly setters?

Only if they can explicitly cause compilation errors, as otherwise it might lead to bugs due to inconsistent behaviour.

```kotlin
val vector = getVector(1f, 1f)
vector.set(2f, 3f)
// Vector2: changes internal state, vector stores 2f, 3f.
// ImmutableVector2: warning; returns a new object, vector still stores 1f, 1f.
```

 > Should we rename them to avoid confusion?

Yes, for the exact same reason.

math/src/main/kotlin/ktx/math/ImmutableVector.kt
jcornaz: >you can provide your own implementation

This would indeed be a very good reason to make this/theses function member of the interface. I wasn't able to think of a specific implementation and making it a member function in the future will be possible without breaking API. (members always have precedence over extension functions).

But I tell all that only to share my thinking. I don't think there's anything to argue. I will, of course, align to your preferences. Shall I make it a member then? If yes, all the functions or only the properties?

czyzby:
As for the inlining, I'll leave it up to you. As I've mentioned in one of the comments, I don't have a strong opinion on the topic. In most cases it would be a premature optimization if you're not using lambdas/reified generics - the warning is out there for a reason - but then again, vectors are pretty low level.

math/src/main/kotlin/ktx/math/ImmutableVector.kt
czyzby: There are other ways of caching results from vectors' methods and I think `ImmutableVector2` currently cannot be extended either way. I don't see how most of these operations could need a different implementation that returns a different value either (e.g. length).

It's a tradeoff between keeping the API close to LibGDX (huge interface), and clean/idiomatic. I'll accept the PR with either approach, just go through all of the methods and consider if any of them might need to be more generic or extendable:

> (...) consider if any of the properties/methods you introduced outside of the interface or class definitions might need overriding.

jcornaz:
Thanks for your answers.

> > Should we rename them to avoid confusion?

> Yes, for the exact same reason.

Any idea of good naming scheme? Shall we prefix them with `get`. Like `getLerp` instead of `lerp`? 

> the warning is out there for a reason

Yes, it is here because `inline` cause the function to expose its implementation, which is considered very bad practice. Modifying the function in the future will then break binary compatibility.

>I'll leave it up to you [...] In most cases it would be a premature optimization

Ok. I will remove *some* inline. I'll leave the ones that are really clearly nothing else than alias, and I'll remove inline modifiers when it is less clear: 
```kotlin

// This is definitely an alias, and deserve to be inlined. There's no arm to expose implementation here.
inline infix fun ImmutableVector2.x(other: ImmutableVector2) = crs(other)

// Here it is debatable. So I'll remove the `inline` modifier
fun <T : ImmutableVector<T>> T.isCollinear(other: T, epsilon: Float = MathUtils.FLOAT_ROUNDING_ERROR): Boolean =
        isOnLine(other, epsilon) && hasSameDirection(other)
```

czyzby:
> Any idea of good naming scheme? Shall we prefix them with `get`. Like `getLerp` instead of `lerp`? 

Sounds good.

> I'll leave the ones that are really clearly nothing else than alias

I think this is the best approach, thanks.

math/src/main/kotlin/ktx/math/ImmutableVector.kt
jcornaz: I agree with everything you said. I'll go through all of them and promote some of them to be member (like `len`).

As a side note I observed that a lot of Java libraries choose to provide "Huge interfaces" as a trade-off to not have to use helper classes, which is unpleasant for the user (this is the case of `Observable` in Rx for instance). But we don't have this limitation in Kotlin :-p)

math/src/test/kotlin/ktx/math/ImmutableVector2Test.kt
```diff
@@ -0,0 +1,701 @@
...
+        assertEquals(ImmutableVector2.ZERO.hashCode(), ImmutableVector2.ZERO.hashCode())
+        assertEquals(ImmutableVector2(-3f, 42f).hashCode(), ImmutableVector2(-3f, 42f).hashCode())
+
+        vectors.forEach {
+            assertEquals(it.hashCode(), it.hashCode())
+        }
+    }
+
+    @Test
+    fun `equals should return true for equivalent vector`() {
```
czyzby: Wrong test name.

math/src/test/kotlin/ktx/math/ImmutableVector2Test.kt
```diff
@@ -0,0 +1,701 @@
...
+
+        vector--
+
+        assertEquals(9f, vector.x, MathUtils.FLOAT_ROUNDING_ERROR)
+        assertEquals(9f, vector.y, MathUtils.FLOAT_ROUNDING_ERROR)
+    }
+
+    @Test
+    fun `should divide vectors by float scalars with div operator`() {
+        val vector = Vector2(10f, 10f)
```
czyzby: Wrong type?

math/src/main/kotlin/ktx/math/ImmutableVector.kt
jcornaz: I changed for the folowing:

```kotlin
override val nor: ImmutableVector2 get() = withLength2(1f)

override fun withLength2(length2: Float): ImmutableVector2 {
  val oldLen2 = this.len2

  return if (oldLen2 == 0f || oldLen2 == length2) this else times(sqrt(length2 / oldLen2))
}
```

Is that ok?

math/src/main/kotlin/ktx/math/ImmutableVector.kt
```diff
@@ -0,0 +1,181 @@
...
+
+    /** Returns true if this vector is in line with the other vector (either in the same or the opposite direction) */
+    fun isOnLine(other: T, epsilon: Float = MathUtils.FLOAT_ROUNDING_ERROR): Boolean
+
+    /** Compares this vector with the other vector, using the supplied epsilon for fuzzy equality testing */
+    fun epsilonEquals(other: T, epsilon: Float): Boolean
+
+    override fun compareTo(other: T): Int = len2.compareTo(other.len2)
+
+    @Deprecated(MUTABLE_METHOD_DEPRECATION_MESSAGE, ReplaceWith("withLerp(target, alpha)"))
```
czyzby: `level = DeprecationLevel.ERROR`

Better play it safe and be a little annoying to the users than cause potential bugs.

czyzby: Same goes for all mutating methods.

jcornaz:
I addressed the mentionned issue.

In a nutshell:
* Few extensions functions are now members of the interface.
* Few functions are no longer inlined
* I renamed obscur method argument names
* I added almost all mutation methods of LibGDX (deprecated)
  * Note that the only documentation is the deprecation message and their argument names are the one in LibGDX
* I prefixed with "with" all functions which were equivalent of a mutation method. 
  * `withLerp`
  * `withInterpolation`
  * `withRotation`
  * etc.

I will soon add usage examples in the README.

math/src/main/kotlin/ktx/math/ImmutableVector.kt
jcornaz: Yes, indeed.

jcornaz:
Please tell me what you think about the examples I added in the readme.

math/src/main/kotlin/ktx/math/ImmutableVector.kt
```diff
@@ -0,0 +1,181 @@
...
+package ktx.math
+
+import com.badlogic.gdx.math.Interpolation
+import com.badlogic.gdx.math.MathUtils
+import com.badlogic.gdx.math.Vector
+import kotlin.math.abs
+import kotlin.math.sqrt
+
+/**
+ * Represent an immutable vector.
```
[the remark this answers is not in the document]
czyzby: Mentioning how you can switch between the two (toMutable/toImmutable) also makes sense.

math/src/main/kotlin/ktx/math/ImmutableVector.kt
```diff
@@ -0,0 +1,181 @@
...
+    /** Returns the opposite vector of same length */
+    operator fun unaryMinus(): T
+
+    /** Returns the result of subtracting the [other] vector from this vector */
+    operator fun minus(other: T): T
+
+    /** Returns the result of adding the [other] vector to this vector */
+    operator fun plus(other: T): T
+
+    /** Returns the same vector with all members incremented by 1 */
```
czyzby: > Returns the same vector

`Returns a new vector instance`

Please update all docs that mention returning "the same" or "this" vector, as this is misleading.

math/src/main/kotlin/ktx/math/ImmutableVector.kt
```diff
@@ -0,0 +1,181 @@
...
+
+/** Returns whether this vector has similar direction compared to the other vector. */
+fun <T : ImmutableVector<T>> T.hasSameDirection(other: T): Boolean =
+        dot(other) > 0f
+
+/** Returns whether this vector has opposite direction compared to the other vector. */
+fun <T : ImmutableVector<T>> T.hasOppositeDirection(other: T): Boolean =
+        dot(other) < 0f
+
+internal const val MUTABLE_METHOD_DEPRECATION_MESSAGE = "Unlike its equivalent in LibGDX, this function doesn't mutate the vector"
```
czyzby: "Mutate" might not be obvious for programming beginners. I'd go with something along the lines of:

`Unlike its equivalent in LibGDX, this function does not change the internal state of the vector and returns a new instance instead. This might break existing code designed with mutable vectors in mind.`

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,209 @@
...
+    override fun unaryMinus(): ImmutableVector2 = ImmutableVector2(-x, -y)
+
+    override operator fun minus(other: ImmutableVector2): ImmutableVector2 = minus(other.x, other.y)
+
+    /** Returns the result of subtracting the given vector from this vector */
+    fun minus(deltaX: Float = 0f, deltaY: Float = 0f): ImmutableVector2 = ImmutableVector2(x - deltaX, y - deltaY)
+
+    override operator fun plus(other: ImmutableVector2): ImmutableVector2 = plus(other.x, other.y)
+
+    /** Returns the result of adding the given vector from this vector */
```
czyzby: `from` -> `to`

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,209 @@
...
+    fun plus(deltaX: Float = 0f, deltaY: Float = 0f): ImmutableVector2 = ImmutableVector2(x + deltaX, y + deltaY)
+
+    override fun inc(): ImmutableVector2 = copy(x = x + 1, y = y + 1)
+    override fun dec(): ImmutableVector2 = copy(x = x - 1, y = y - 1)
```
czyzby: I'd keep the code consistent. If you're not using `copy` for other methods like `plus`, continue using the constructor here as well.

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,209 @@
...
+                y = y * invAlpha + target.y * alpha
+        )
+    }
+
+    override fun withRandomDirection(): ImmutableVector2 = withAngleRad(MathUtils.random(0f, MathUtils.PI2))
+
+    override fun epsilonEquals(other: ImmutableVector2, epsilon: Float): Boolean =
+            epsilonEquals(other.x, other.y, epsilon)
+
+    /** Compares this vector with the other vector, using the supplied epsilon for fuzzy equality testing */
```
czyzby: Would you mind adding parameters documentation? `epsilon` might not be obvious for everyone.

math/src/test/kotlin/ktx/math/ImmutableVector2Test.kt
```diff
@@ -0,0 +1,712 @@
...
+
+    @Test
+    fun `should divide vectors by float scalars with div operator`() {
+        val vector = ImmutableVector2(10f, 10f)
+
+        val result = vector / 2.5f
+
+        assertEquals(4f, result.x, MathUtils.FLOAT_ROUNDING_ERROR)
+        assertEquals(4f, result.y, MathUtils.FLOAT_ROUNDING_ERROR)
+    }
```
czyzby: Most of these tests check if the implementation is the same as the official one, but not if it is correct. I'm not saying there are any bugs in the code, but if there were present in both our and LibGDX code, these tests wouldn't expose that at all. You've already done a ton of work on this PR, and this is not a pressing issue, but if you don't mind adding a few tests that check the results of `ImmutableVector2` methods against actual pre-calculated values that we know are correct, I'd be much more confident before publishing this code. You don't have to do it for every element of `vectors` though.

math/README.md
```diff
@@ -39,6 +39,72 @@ Note that since `Shape2D` has `contains(Vector2)` method, `in` operator can be u
...
+mutation methods return new vectors instead of mutate the reference.
+- Note that one may want to create type aliases to makes the usage more concise: `typealias Vect2 = ImmutableVector2`
+- `ImmutableVector` is comparable (`>`, `>=`, `<`, `<=` are available). Comparison is evaluated by length
+- instances can be destructed: `val (x, y) = vector2`
+- `Vector2.toImmutableVector2()` Returns an immutable vector with same `x` and `y` attributes than this `Vector2`
+- `ImmutableVector2.toVector2()` Returns an mutable vector with same `x` and `y` attributes than this `ImmutableVector2`
+- Most of the functions of `Vector2` which mutate the vector are provided but deprecated. This allow smooth migration from
+`Vector2`. 
+- Notable difference with `Vector2`
+  - `+`, `-`, `*`, `/` are available and replace `add`, `sub` and `scl`.
```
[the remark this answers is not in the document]
jcornaz: Mmmh.  But it is the case. Isn't it?

math/README.md
```diff
@@ -39,6 +39,72 @@ Note that since `Shape2D` has `contains(Vector2)` method, `in` operator can be u
...
+
+Creating typealias
+```kotlin
+import ktx.math.*
+
+typealias Vec2 = ImmutableVector2
+
+var v1 = (Vec2.X + Vect2.Y).nor
+var v2 = Vec2(1f, 2f).withLength(3f)
+```
```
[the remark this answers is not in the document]
jcornaz: Should I commit that file? In tests maybe.

math/README.md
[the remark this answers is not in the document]
czyzby: Oh, I see. The README kind of suggests they are completely removed. Nevermind.

math/README.md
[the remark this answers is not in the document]
czyzby: It's not necessary, just a double check if there are anymore typo or missing imports.

math/src/test/kotlin/ktx/math/ImmutableVector2Test.kt
jcornaz: This is a fair concern and I share it, especially after finding incorrect behavior in LibGDX's vector.

And because some implementation also delegate to LibGDX, a bug in LibGDX would necessirly be present and not detected.

I'll try to add some check against actual pre-calculated values.

czyzby: Thanks, much appreciated.

math/README.md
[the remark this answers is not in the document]
jcornaz: Yes, the README kind of suggests that they are not available at all.... Which is somehow true, since deprecation level is "error", and the user cannot use them.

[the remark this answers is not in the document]
czyzby: Agreed, this is OK. I've written this comment before going thoroughly through the code, since the README shows up on the top of the page.

jcornaz:
Quick update to say I didn't forget this PR. I just lack of free time at the moment. I'll try to address the remaining issues next weekend.

czyzby:
@jcornaz It's OK, you've already done a lot, and I'm also pretty busy recently. I could take over the extra tests, but I'm not sure if I'll be able to make it this week. Just let me know if you won't have time to finish it after all.

jcornaz:
I finally addressed the request changes.

In a nutshell:
* I fixed typos and improve documentation as requested
* I tested the examples in readme in order to verify their correctness
* I wrote a bunch of test in order to make sure that it works as expected even if there is a bug in LibGDX.

While writing the new tests, I actually noticed a problem in LibGDX vector when it comes to orientation of the vector zero and with many orientation comparison (`hasSameDirection`, `isOnLine`, `isCollinear`, etc.)

LibGDX is not consistent accross theses functions when the vector zero is involved. I opted for correctness instead of replicating errors of LibGDX. The orientation of a zero vector is *undefined* therefore:

* angles function involving one or two zero vectors return `NaN`. (instead of an arbitrary angle, like LibGDX does)
* `isOnline` return false if one of the vectors is zero. (instead of returning true, like LibGDX does)

Other direction-related functions (`isCollinear`, `isPerpendicuar`, etc.) have not been touched, since LibGDX was already behaving correctly for them.

czyzby:
I'll review the changes and update KTX to LibGDX 1.9.9 in the week after this one after I'm done with a conference and a hackathon. Thanks again for the contribution!

jcornaz:
Note this PR contains too many commits. I'd recommend to squash before merge. (Either by choosing squash-merge in github, or I may rebase if you prefer)

czyzby:
But 36 commits would move you to the top of the contribution statistics. ;)

I honestly don't mind either way - the extra commits make it easier to see how the code progressed and normally I'm a fan of "commit early, commit often". Squashed commits make it easier to create changelogs based on the history, but then again - we usually do it manually as the changes are added either way.

jcornaz:
It is as you prefer.

I like to "commit early, commit often" when working on a feature  (which explains my 36 commits^^). But once the feature is merged I no longer care that It took x commits to be written. I only care about the changes made by the feature as whole. And the original commits are never lost with squash-commit anyway. They stay attached to the PR, and may still be anylized independently if needed (but I never came back the specific commit of a PR personally).

But at the end it's up to you, of course. ;-)

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,218 @@
...
+     *
+     * This method is faster than [dst] because it avoids calculating a square root. It is useful for comparisons,
+     * but not for getting exact distance, as the return value is the square of the actual distance.
+     */
+    fun dst2(otherX: Float, otherY: Float): Float = Vector2.dst2(x, y, otherX, otherY)
+
+    /** @return the distance between this and the other vector */
+    fun dst(otherX: Float, otherY: Float): Float = Vector2.dst(x, y, otherX, otherY)
+
+    /** Returns the result of multiplying this vector by the given matrix */
```
czyzby: I'd include a comment about the assumptions that this method makes. You cannot normally multiply a 2D vectors with a 3D matrix, can you?

math/src/main/kotlin/ktx/math/ImmutableVector.kt
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * Returns the squared distance between this and the other vector
+     *
+     * This method is faster than [dst] because it avoids calculating a square root. It is useful for comparisons,
+     * but not for getting exact distance, as the return value is the square of the actual distance.
+     */
+    infix fun dst2(vector: T): Float
+
+    /** Linearly interpolates between this vector and the target vector by alpha */
```
czyzby: It seems that the current Kotlin convention proposes to include `[references]` to parameters in the doc comments. Using `@param` would be excessive, but I'd appreciate if you wrapped at least one reference to each parameter mentioned in each doc comment with a link. If you don't have the time, this is OK - it's not crucial, I can add it after merging.

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
```diff
@@ -0,0 +1,218 @@
...
+
+import com.badlogic.gdx.math.MathUtils
+import com.badlogic.gdx.math.Matrix3
+import com.badlogic.gdx.math.Vector2
+import kotlin.math.*
+
+/**
+ * Represent an immutable vector 2D
+ *
+ * You may use [ImmutableVector2.toMutable] to crate a (mutable) [Vector2] from a an [ImmutableVector2].
```
czyzby: `crate` -> `create`.

math/src/test/kotlin/ktx/math/ImmutableVector2Test.kt
```diff
@@ -0,0 +1,1436 @@
...
+        assertEquals(0f, ImmutableVector2.X.dot(0f, 1f))
+
+        assertEquals(0f, ImmutableVector2.X.dot(0f, 0f))
+        assertEquals(0f, ImmutableVector2.X.dot(0f, 0f))
+
+        assertEquals(2f, ImmutableVector2(1f, 1f).dot(1f, 1f))
+        assertEquals(2f, ImmutableVector2(1f, 1f).dot(1f, 1f))
+
+        assertEquals(7f, ImmutableVector2(-2f, 3f).dot(4f, 5f))
+        assertEquals(7f, ImmutableVector2(-2f, 3f).dot(4f, 5f))
```
czyzby: Why are you invoking these methods twice with the same parameters? Did you want to compare it to the mutable vector results or test both dot functions (vector/2 floats)?

math/src/test/kotlin/ktx/math/ImmutableVector2Test.kt
```diff
@@ -0,0 +1,1436 @@
...
+        }
+    }
+
+    @Test
+    fun `withRandomDirection should return a variety of random vectors`() {
+        val vectors = HashSet<ImmutableVector2>()
+        repeat(1000) {
+            vectors += ImmutableVector2.X.withRandomDirection()
+        }
+        assertTrue(vectors.size > 900)
```
czyzby: Use a fixed random seed. :)

math/src/test/kotlin/ktx/math/ImmutableVector2Test.kt
```diff
@@ -0,0 +1,1436 @@
...
+    fun `isOnline should be commutative`() {
+        vectors.forEach { v1 ->
+            vectors.forEach { v2 ->
+                assertEquals(v1.isOnLine(v2), v2.isOnLine(v1))
+            }
+        }
+    }
+
+    @Test
+    fun `isCollinear should return same than Vector2`() {
```
czyzby: `should return same value as`

math/src/main/kotlin/ktx/math/ImmutableVector2.kt
jcornaz: is `apply the given affine transformation and return the resulting vector` better`? It would make  clearer that we assume that the matrix is an `affine transformation` representation.

czyzby: I meant that this method assumes the vector has a `z=1` component during the multiplication (hence the addition of unmodified `val[6]` and `val[7]`) and "removes" the last element after the operation. This might not be obvious, especially for someone who does not deal with graphics very often.

jcornaz: >this method assumes the vector has a z=1 component

Yes, because it assumes it is an *affine transformation*.

>This might not be obvious, especially for someone who does not deal with graphics very often. 

Yes. But I'm a bit affraid of making it look complicated if we talk about `z=1`, wherever if we only took about *affine transformation*, there is no need for the user to understand how they work. he only need to pass a valid affine transformation, which he can get from LibGDX.

Actually, it makes me think we may change the signature for `operator times(transform: Affine2): Vector2`. It would make it clearer about what the contract is. And experienced user can easily create an `Affine2` out of a `Matrix3`.

jcornaz: And I may also rewrite the test for something more meaningful.

For instance:

```kotlin
val vector = ImmutableVector(3f, 4f)

transformation = Affine2()
transformation.preTranslate(-2f, 3f)
transformation.preRotate(18f)
transformation.preScale(1.5f, 2f)

assertEquals(vector.plus(-2f, 3f).withRotation(18f).scl(1.5f, 2f), vector * transformation)
```

czyzby: > Yes, because it is how *affine transformation* matrices are defined.

Which isn't mentioned in the documentation or the method name, hence the comment. Method name suggests a simple matrix multiplication, in which case the dimensions simply would not match. I think mentioning both *how* and *why* is worth including in the doc.

jcornaz: >Which isn't mentioned in the documentation or the method name, hence the comment. 

Yes, of course, it should be added. I'm not challenging your initial comment. I also think it deserve to be clarified.

To be clear, I propose to change the documentation and signature to the following:

```kotlin
/** Apply the given affine [transformation] and return the resulting vector */
operator fun times(transformation: Affine2): ImmutableVector2
``` 

Here method signature and comment would clearly state that an affine transformation is needed and that the function returns the result of applying the transformation to the vector.

Whould that be ok for you?

czyzby: I like this approach. The original LibGDX `Vector2` has the misleading `mul(Matrix3)`, so it's a matter of compatibility vs usability vs clarity of API. I think we can add a deprecated `mul(Matrix3)` method (since the original mutates) that points to `times(Affine2)` - `Affine2` is fairly easy to obtain with a `Matrix3` either way.

czyzby:
I think this is among the biggest contributions to KTX. Thanks for your work and putting up with my nitpicking. :)

Issue #162, opened 2018-10-02
ShapeRenderer.use


keturn:
The GDX ShapeRenderer has `begin` and `end` methods. Should we make the ktx-graphics version have a `use` block, the same way Batch and ShaderProgram do?

czyzby:
Sure. We just make sure to add the `ShapeType` parameter. Do you want to add this utility in a PR or should I implement it?

keturn:
Yeah, I expect I'll be able to make a PR for this.

czyzby:
Thanks for the PR, @keturn.

PR #163, opened 2018-10-04
ShapeRenderer.use()

keturn:
Addresses #162 

The test passes, but since the test runs on a mock, I must confess I haven't yet run this live. (I haven't looked up yet how to link my project against a local fork of libktx instead of using the one from the repo.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 #### 1.9.8-SNAPSHOT
 
+- **[FEATURE]** (`ktx-graphics`) Added `ShapeRenderer.use` to allow omission of `begin()` and `end()` calls.
+
 #### 1.9.8-b5
 
 - **[UPDATE]** Updated to Kotlin 1.2.70.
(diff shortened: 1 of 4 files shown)


czyzby:
> I haven't looked up yet how to link my project against a local fork of libktx instead of using the one from the repo.

[Working from sources guide](https://github.com/libktx/ktx/blob/master/.github/CONTRIBUTING.md#working-from-sources) might be useful. When you run `gradle build install`, it puts the libraries in your Maven Local, so if your LibGDX Gradle project defines `mavenLocal()` repository over snapshots repo, it will use your local library versions instead. Just make sure to use a snapshot release, such as `1.9.8-SNAPSHOT`.

> I must confess I haven't yet run this live.

It's OK, this is a very simple utility and tests should be sufficient. Unless the renderer itself is working incorrectly, this function shouldn't cause any issues.

@keturn, could you also update the changelog file?

keturn:
Thanks for the pointers for local install.

I succeeded in running this code locally and added an entry in the changelog.

czyzby:
Thanks. Expect a snapshot release with your changes soon.

Issue #154, opened 2018-09-03
Rewrite async modules to non-experimental coroutines


czyzby:
Experimental coroutine APIs are deprecated and yield errors in Kotlin 1.3.

czyzby:
`ktx-async` uses some APIs deprecated as of `0.26.1`.

lokeshj:
@czyzby Now that coroutines are 1.0, could you please take a look at this again?

jcornaz:
`ktxAsync` should be an extension on `CoroutineScope` as well to be consistant with other coroutine builders of `kotlinx.coroutines` and make life-cycle mangements easier.

czyzby:
The `develop` branch is compatible with `1.0.1` coroutines API as of merging #170, although I still think `ktx-async` should be rewritten to leverage new coroutine APIs. See #169.

jcornaz:
I know it is likely to be solved when you'll rewrite the `async` module, but I'd like to report here a bug I've fount using 3.9.9-SNAPSHOT today: `AssetStorage.load` is not thread safe.

Example:
```kotlin
ktxAsync {
  val texture1 = async { assetStoreage.load<Texture>("texture1.png") }
  val texture2 = async { assetStoreage.load<Texture>("texture2.png") }

  addScreen(MyScreen(texture1.await(), texture2.await())
  setScreen<MyScreen>()
}
```

With the code above, I experienced visual glitch like inverted textures, or getting only partial or scaled texture. Removing usages of `async` solved the problem.

Thread safety is an implicit contract of any `suspend` function. If necessary the function may use a `Mutex.withLock`, `withContext` or channels to achieve its thread safety. But I always expect from a suspending function to be callable concurrently from any thread.

czyzby:
I know, the new solution uses a single `assets` mutable collection protected by a `Mutex`.

czyzby:
Just wanted to let you know that I'm currently dealing with work and the end of semester at my uni, so I had to postpone this "a little". I should be done in a few weeks.

czyzby:
@jcornaz I finished the documentation of the new `ktx-async` module and pushed the overhaul to the `develop` branch. I decided to remove the `AssetStorage`, since I don't think it should be strictly connected with the core coroutines utilities and I have to admit I'm not done with that one yet. [Could you take a look at the new API](https://github.com/libktx/ktx/tree/develop/async)?

czyzby:
I'm closing this issue for now, `AssetStorage` will be implemented separately.

Issue #169, opened 2018-12-04
LibGdx main dispatcher


jcornaz:
Hi,

In my humble opinion, the `async` module misses the only feature I'd expect him to provide: A dispatcher for the LibGDX main thread.

Here is how to write it:

```kotlin
public val Dispatchers.LibGdx: CoroutineDispatcher get() = LibGdxDispatcher

private object LibGdxDispatcher : MainCoroutineDispatcher() {

  @ExperimentalCoroutinesApi
  override val immediate = object : MainCoroutineDispatcher() {
    @ExperimentalCoroutinesApi
    override val immediate: MainCoroutineDispatcher get() = this

    override fun dispatch(context: CoroutineContext, block: Runnable) = block.run()
  }

  override fun dispatch(context: CoroutineContext, block: Runnable) = Gdx.app.postRunnable(block)
}
```

And how to use it:

```kotlin
fun CoroutineScope.startLibGdxCoroutine() = launch(Dispatchers.Main) { // <-- We can use any coroutine builder

  // Here we are on a frame update by the LibGDX thread.
  
  delay(1000) // <-- Call any suspending function (no need to use the `KtxAsync.delay`)
 
  // Here we are back on a frame update by the LibGDX thread

  withContext(Dispatchers.Main) {
    // It is not needed in this case, but one may use `withContext` to perform an operation on a frame update of the LibGDX thread
  }
}
```

czyzby:
Already on this. New coroutines API requires a complete redesign of the `ktx-async` module. I cannot give you any time estimates, but just wanted you to know I'm getting close to the next release.

There will be 2 dispatchers based on LibGDX API: for the rendering thread (using `Gdx.app.postRunnable`) and wrapping `AsyncExecutor` (using `executor.submit`).

jcornaz:
Thank you very much for your work. Game programming is very pleasant with Kotlin and KTX ;-)

czyzby:
I'm going to close this as a duplicate of #154, but I'm still working on this. Recent weeks were pretty busy for me, but now I have a bit more free time - I want to finish the `ktx-async` refactor and immutable vectors merge by the end of the week.

PR #170, opened 2018-12-04
Update to kotlin coroutines 1.0.1

ProOrange:
https://github.com/libktx/ktx/issues/154


czyzby:
Thanks for your help. This is a good first step towards `1.3` compatibility, although I think ultimately `ktx-async` should be rewritten to use the new coroutine APIs. See #169.

PR #174, opened 2018-12-08
`+` Operator for creating sequence of actions

### LINKED ISSUE #172: `+` operator to create sequence of actions
I think a `+` operator to easily create a sequence of actions would be nice

Example:
```kotlin
val sequence = moveTo(200f, 200f, 0.5f) + rotateTo(90f, 0.5f) + fadeOut(0.5f) + remove()
```

jcornaz:
Resolves #172 

As I proposed in #172, I also made `SequenceAction.then(Action)` consistant with `Action.then(Action)` by never mutating the reciever.

### DIFF
--- a/actors/src/main/kotlin/ktx/actors/actions.kt
+++ b/actors/src/main/kotlin/ktx/actors/actions.kt
@@ -49,7 +49,7 @@ infix fun Action.then(action: Action): SequenceAction {
 
 private fun SequenceAction.addUnwrapped(action: Action) {
   if (action is SequenceAction) {
-    action.actions.forEach(this::addAction)
+    action.actions.forEach(::addAction)
   } else {
     addAction(action)
   }
--- a/actors/src/test/kotlin/kts/actors/actionsTest.kt
+++ b/actors/src/test/kotlin/kts/actors/actionsTest.kt
@@ -59,7 +59,7 @@ class ActionsTest {
   }
 
   @Test
-  fun `action then action should chain actions into sequences`() {
+  fun `should chain actions into sequence with action then action`() {
     val firstAction = MockAction()
     val secondAction = MockAction()
 
@@ -71,7 +71,7 @@ class ActionsTest {
   }
 
   @Test
-  fun `sequence then action should chain underling actions into sequence`() {
+  fun `should chain underling actions into sequence with sequence then action`() {
     val firstAction = MockAction()
     val secondAction = MockAction()
     val thirdAction = MockAction()
@@ -85,7 +85,7 @@ class ActionsTest {
   }
 
   @Test
-  fun `action then sequence should chain underling actions into sequence`() {
+  fun `should chain underling actions into sequence with action then sequence`() {
     val firstAction = MockAction()
     val secondAction = MockAction()
     val thirdAction = MockAction()
@@ -99,7 +99,7 @@ class ActionsTest {
   }
 
   @Test
-  fun `sequence then sequence should chain underling actions into sequence`() {
+  fun `should chain underling actions into sequence with sequence then sequence`() {
     val firstAction = MockAction()
     val secondAction = MockAction()
     val thirdAction = MockAction()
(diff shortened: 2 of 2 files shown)


jcornaz:
Ok, I updated the code (and tests) and it nows always extract action from sequence.

actors/src/main/kotlin/ktx/actors/actions.kt
```diff
@@ -34,24 +34,25 @@ operator fun Actor.minusAssign(action: Action) = removeAction(action)
...
+infix fun Action.then(action: Action): SequenceAction {
+  val result = SequenceAction()
+  result.addUnwrapped(this)
+  result.addUnwrapped(action)
+  return  result
+}
+
+private fun SequenceAction.addUnwrapped(action: Action) {
+  if (action is SequenceAction) {
+    action.actions.forEach(this::addAction)
```
czyzby: Isn't `this` optional in this case?

To you think we should unwrap the actions recursively (`forEach(::addUnwrapped)`)?

actors/src/test/kotlin/kts/actors/actionsTest.kt
```diff
@@ -58,7 +59,7 @@ class ActionsTest {
   }
 
   @Test
-  fun `should chain actions into sequences`() {
+  fun `action then action should chain actions into sequences`() {
```
czyzby: Personal preference: tests names starting with `should`. E.g.: `should chain actions into sequences with then infix function`.

actors/src/test/kotlin/kts/actors/actionsTest.kt
```diff
@@ -85,44 +85,47 @@ class ActionsTest {
...
     assertEquals(secondAction, sequence.actions[1])
-    assertEquals(2, sequence.actions.size)
+    assertEquals(thirdAction, sequence.actions[2])
+    assertEquals(fourthAction, sequence.actions[3])
+    assertEquals(4, sequence.actions.size)
   }
 
   @Test
-  fun `should add multiple actions into sequences`() {
+  fun `should not mutate multiple actions with then`() {
```
czyzby: Not sure if this test is necessary. You could add extra asserts to existing tests with sequences.

actors/src/test/kotlin/kts/actors/actionsTest.kt
```diff
@@ -131,12 +134,28 @@ class ActionsTest {
...
+    assertEquals(firstAction, sequence.actions[0])
+    assertEquals(secondAction, sequence.actions[1])
+  }
+
+  @Test
+  fun `should add action to sequence with +=`() {
+    val firstAction = MockAction()
+    val secondAction = MockAction()
+    val thirdAction = MockAction()
+
```
czyzby: Remove empty line.

actors/src/main/kotlin/ktx/actors/actions.kt
jcornaz: Ah yes

actors/src/test/kotlin/kts/actors/actionsTest.kt
jcornaz: This would mix the responsibilty of the tests IMHO. One test make sure one can create a sequence with `then`. And one test make sure that `then` do not mutate (anymore).

But it is personal perference. Please confirm if you want me to change it.

actors/src/main/kotlin/ktx/actors/actions.kt
czyzby: For the record - extracting actions recursively might be excessive and rare in practice.

actors/src/test/kotlin/kts/actors/actionsTest.kt
czyzby: No, this is OK.

actors/src/main/kotlin/ktx/actors/actions.kt
jcornaz: > extracting actions recursively might be excessive and rare in practice.


Yes, which is why I didn't do it in the first place. 

Actually I said "Ah yes" about the "this" being optional.

I'm not a big fan of doing `addUnwrapped` recursively, for the reason you mentionned. Please confirm if you want me to make it recursive.

czyzby:
Thanks.

PR #186, opened 2019-04-15
Type-safe enum names without invoke() method and brackets

FocusPo1nt:
Example:

enum class Drawables {
buttonUp,
buttonDown,
buttonChecked;
}

skin(myAtlas) {
button {
up = it[buttonUp]
down = it[buttonDown]
}
}


czyzby:
I wanted to avoid enum utilities, since they are can be very application-specific, but one actually seems pretty generic. Thanks!

Could you write some simple unit tests for your utilities? You can see some examples [in this file](https://github.com/libktx/ktx/blob/master/style/src/test/kotlin/ktx/style/styleTest.kt).

FocusPo1nt:
I added the test.

From Style documentation: "..we encourage you to store your drawables, fonts, colors and non-default styles names as enums to ensure complete safely at compile time."

I like this approach. I just wanted to make it more readable a little bit

czyzby:
Yeah, it's a very good idea. I'll update the README as well.

czyzby:
Thanks.

PR #187, opened 2019-05-14
Removed extra spaces in default log tags

maltaisn:
I don't think these spaces were meant to be there, but I don't know. Now all tags are consistent though.
It currently looks like this when logging an info message:

    [[INFO] ] This is my info message


czyzby:
It was deliberate to achieve the same indentation level for all logs. I agree that it could have been achieved by proper log format (same for the parenthesis), but it's so small of a change that I'd keep it as is for consistency with the older versions.

PR #189, opened 2019-06-02
Added default parameter value for Skin.get

### LINKED ISSUE #188: Default parameter value for Skin.get
The skin module has an extension method to get an asset by name, with reified type.
```kotlin
val style = skin.get<ImageButtonStyle>("custom-style")
```
When using the "default" named asset, we have two choices:
```kotlin
val style = skin.get<ImageButtonStyle>("default")
val style: ImageButtonStyle = skin.get(ImageButtonStyle::class.java)
```
Ideally `Skin.get` would have a default value of `"default"` for the name parameter, allowing:
```kotlin
val style = skin.get<ImageButtonStyle>()
```
Can this be added?

maltaisn:
Closes #188 

`Skin.get` was an infix method. I'm not sure how useful that is... why would someone choose the infix syntax over the bracket syntax? It prevented adding default parameter values so I removed it and update the tests and changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 #### 1.9.9-SNAPSHOT
 
 - **[UPDATE]** Updated to Kotlin 1.3.21.
+- **[CHANGE]** (`ktx-style`) `Skin.get` is no longer infix to allow a default parameter.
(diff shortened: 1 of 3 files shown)


czyzby:
Agreed, `infix` operator was excessive. Thanks for the PR.

PR #194, opened 2019-06-09
Added json extensions module

maltaisn:
Improves syntax with type inference and improves type safety of LibGDX Json reader and writer. The module doesn't add anything new, just reified extension function.

I didn't do the readme yet because I realized this module doesn't add any functionality apart from syntax, unlike other modules. For the same reason, only minimal tests were made.

### DIFF
--- a/json/README.md
+++ b/json/README.md
@@ -1,0 +1,1 @@
+# KTX: Json extensions
--- a/json/README.md
+++ b/json/README.md
@@ -1,1 +1,80 @@
 # KTX: Json extensions
+
+Extension methods for LibGDX JSON reader and writer.
+
+### Why?
+
+The JSON reader and writer in LibGDX uses a lot of `Class` parameter, which leads to a lot
+of `Type::class.java` in method arguments. Kotlin brings reified generics which create the
+possibility to get the `Class` of a generic type parameter. This module offers extension methods 
+with reified generics to avoid using `::class.java` in your code and to allow type inference and
+in some cases better type safety.
+
+### Guide
+A JSON instance can be used the same as usual, but using the extension method whenever there's
+a class parameter in the original method.
+
+##### Usage examples
+
+Creating a new JSON instance with custom parameters:
+```kotlin
+import ktx.json.*
+
+val json = Json()
+
+// Add shorthands for two classes
+json.addClassTag<Vector2>("vec2")
+json.addClassTag<Color>("color")
+
+// Set the type of elements in the "cards" collection of Player objects
+json.setElementType<Player, Card>("cards")
+
+// Add a custom serializer for Vector2
+json.setSerializer(object : Json.Serializer<Vector2>() { /* ... */ })
+```
+
+A class with a custom serializer:
+```kotlin
+import ktx.json.*
+
+class Player : Json.Serializable {
+  lateinit var pos: Vector2
+  lateinit var cards: List<Simple>
+
+  override fun read(json: Json, jsonData: JsonValue) {
+    pos = json.readValue("pos", jsonData)  // Type inference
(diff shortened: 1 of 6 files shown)


czyzby:
Thanks for the PR, I'll review it soon.

Keep in mind that the PRs should be made to `develop` branch rather than `master`. I know it's a bit confusing since the `master` branch is the default one, but this is deliberate - I want people to browse the relevant code that they would find in the current Maven release rather than new features, which might not be included yet.

> I didn't do the readme yet because I realized this module doesn't add any functionality apart from syntax, unlike other modules.

That's not entirely true, `ktx-math` and `ktx-collections` are basically syntax sugar modules and they still list all of the functionalities and show you some practical examples. I'd like to keep this sort of documentation for every module. I could still accept the PR without it, but keep in mind I'll not merge it into `master` and include in the next release before writing such documentation myself. I think the documentation is what made people use and contribute to this library in the first place. :)

> For the same reason, only minimal tests were made.

I agree that you shouldn't test the whole LibGDX JSON API, but please make sure each included extension method is tested by at least one test case.

maltaisn:
> Keep in mind that the PRs should be made to `develop` branch rather than `master`.

If I did it correctly, the commit was made on a new `maltaisn/json` branch, branched from the `maltaisn/develop` branch (which may not be up-to-date with `libktx/develop`). I created a new branch since I intended to do many commits (which I haven't made yet).

> I agree that you shouldn't test the whole LibGDX JSON API, but please make sure each included extension method is tested by at least one test case.

I'll make the tests soon.

I'll also do the readme following the structure of other readmes, and mentioning the fact that other serialization libraries exist and should be preferred in some cases.

czyzby:
You used the correct branch, but made the request to merge into `libktx:master` rather than `libktx:develop`. Don't worry, I changed it already.

maltaisn:
I added some tests and made a readme draft, you can complete it if anything is missing.

czyzby:
There are only minor changes required (e.g. missing trailing new lines in some files), but I can correct these manually. Thanks for the contribution.

czyzby:
Somehow passing `name: String, jsonData: JsonValue` in that order feels wrong. I'd include `jsonData` as the first parameter in most methods, but I understand consistency with LibGDX API might be more important in this case.

czyzby:
Not sure if you were aware of this, but you can use Kotlin data classes along with LibGDX `Json` API as long as you manage to provide defaults for each variable. You can check the latest commits to develop to see my slight refactoring.

Anyway, a snapshot of the libraries is being uploaded to Maven as I'm writing this. Thanks again for the contribution.

maltaisn:
> Somehow passing `name: String, jsonData: JsonValue` in that order feels wrong.

As you said that was for consistency with the original API but I agree the order feels wrong. In fact, if we put name after, the two `readValue` methods could be merged into a single one with a default argument value of `null` for name. (Same for `readArrayValue`)


> Not sure if you were aware of this, but you can use Kotlin data classes along with LibGDX `Json` API as long as you manage to provide defaults for each variable. You can check the latest commits to develop to see my slight refactoring.

No I didn't know thanks! I just checked the Kotlin documentation:

> Note that, as described in Secondary Constructors, if a class has default values for all constructor parameters, a public no-argument constructor will be generated for it. This works even if the `@JvmOverloads annotation is not specified.

So it should work with all classes not just data classes.

czyzby:
> As you said that was for consistency with the original API but I agree the order feels wrong. In fact, if we put name after, the two `readValue` methods could be merged into a single one with a default argument value of `null` for name. (Same for `readArrayValue`)

Now I'm tempted to do that. Our goal is to improve LibGDX after all.

> So it should work with all classes not just data classes.

Yes, it's just that data classes are the most useful in terms of serialized data objects.

PR #195, opened 2019-06-24
Merged JSON two readValue and readArrayValue functions into one

maltaisn:
As discussed [here](https://github.com/libktx/ktx/pull/194#issuecomment-501152163).

The `name` and `jsonData` parameters were inverted to allow the default value of `name` to be null. When the name is null, the `jsonData` is parsed to type `T`. Otherwise, a named attribute of a `jsonData` (a json map) is parsed to type `T`.


json/src/main/kotlin/ktx/json/json.kt
```diff
@@ -39,24 +39,16 @@ inline fun <reified T, reified E> Json.setElementType(fieldName: String) =
...
+    readValue(T::class.java, if (name == null) jsonData else jsonData.get(name))
 
 /**
  * Read an iterable value of type [T] with elements of type [E] from a [jsonData] attribute with a [name].
+ * If [name] is `null`, this will directly read [jsonData] as an iterable of type [T].
  */
-inline fun <reified T : Iterable<E>, reified E> Json.readArrayValue(name: String, jsonData: JsonValue): T =
-    readValue(name, T::class.java, E::class.java, jsonData)
+inline fun <reified T : Iterable<E>, reified E> Json.readArrayValue(jsonData: JsonValue, name: String? = null): T =
+    readValue(T::class.java, E::class.java, if (name == null) jsonData else jsonData.get(name))
```
czyzby: Missing new line at the end of the file.

PR #203, opened 2019-07-26
adding depth buffer clear to clearScreen for screens with 3d graphics

jojomickymack:
this was branched off of develop instead of master.


czyzby:
Thanks!

Issue #206, opened 2019-08-21
Minimize sequence unwrapping and instantiating


cypherdare:
I started to do a PR for this but it causes an existing test to fail, so I think it needs discussion.

IMO, `SequenceAction.then` should not unwrap its children if it can be avoided. Consider:

    val sequence = moveTo(1f, 2f) + moveTo(3f, 4f) + moveTo(5f, 6f) + moveTo (7f, 8f)

The first operation creates a SequenceAction. The second operation immediately unwraps that sequence and creates a new sequence to put the same two original actions in. Then the third operation unwraps that sequence to transfer three actions.

So a simple sequence of 4 actions results in 3 SequenceActions instantiated and has transferred actions 5 times.

If we do this:

```
infix fun Action.then(action: Action): SequenceAction {
  if (this is SequenceAction) {
      this.addUnwrapped(action)
      return this
    }
  val result = Actions.sequence()
  result.addUnwrapped(this)
  result.addUnwrapped(action)
  return result
}
```

Only one SequenceAction is instantiated and no actions have to be unwrapped, no matter how long the chain of actions is. However, this fails the test `should not mutate multiple actions with then`, which raises the question of, _Why shouldn't it mutate if it can avoid all the extra memory churn?_ Why is it better to throw away a SequenceAction than to mutate it?

Incidentally, I think `addUnwrapped` should also free the SequenceActions that it throws away, since they are no longer useful and their reference is likely to be lost. Like this:

```
private fun SequenceAction.addUnwrapped(action: Action) {
  when (action) {
    is SequenceAction -> {
      action.actions.forEach(::addAction)
      action.pool?.free(action)
    }
    else -> addAction(action)
  }
}
```

czyzby:
I believe that's how it was implemented at first, but we eventually decided against it, since `+` operator in Kotlin should not mutate either of the operands. See #172 #174

cypherdare:
Ah, OK. Then I would suggest we change then() back to the way it was and
replace + with +=. Otherwise, it would be poor practice to ever use either
of these if you don't like your game stuttering.

czyzby:
> Otherwise, it would be poor practice to ever use either of these if you don't like your game stuttering.

Depends on the amount of actions that you use and the target device. Unless you're creating thousands of sequence actions, creating a few extra short-lived objects here and there should not be an issue. Modern JVMs are rather powerful. On desktop this is unlikely to be an issue. On mobile - well, you can deal with it _if_ it becomes an issue.

[...]

PR #207, opened 2019-08-23
SequenceAction and ParallelAction functions refactoring

cypherdare:
An implementation of what we discussed in issue #206.

(Edited post discussion and review)
- All unwrapping removed because it is bug-prone.
- `then` is now mutating if the first action is a SequenceAction. The second action is simply added to it.
- `+` is still non-mutating.
- `parallelTo` deprecated for new `along`
- `along` is mutating if the first action is a ParallelAction. The second action is simply added to it.
- `/` added to make ParallelActions without mutating.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 #### 1.9.10-SNAPSHOT
-- **[CHANGE]** `Action.parallelTo` changed to `Action.along` and it now mutates the caller and argument if they are ParallelActions, to save wasted instantiations and avoid GC.
+- **[CHANGE]** `Action.parallelTo` changed to `Action.along`.
+- **[CHANGE]** `Action.along` (formerly `Action.parallelTo`) and `Action.then` no longer unwrap the second action.
+- **[CHANGE]** `ParallelAction.along` (formerly `ParallelAction.parallelTo`) and `SequenceAction.then` simply add the second action to the group without unwrapping.
 - **[UPDATE]** Added `/` operator to Action, which performs the non-mutating version of along, wrapping the caller and argument in a new ParallelAction.
-- **[CHANGE]** `Action.then` now mutates the caller and argument if they are SequenceActions, to save wasted instantiations and avoid GC. The `+` operator still performs the old non-mutating version, wrapping the caller and argument in a new SequenceAction.
 - **[CHANGE]** `ParallelAction.plus()` and `SequenceAction.plus()` no longer unwrap their components.
 
 #### 1.9.10-b1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 #### 1.9.10-SNAPSHOT
+- **[CHANGE]** `Action.parallelTo` changed to `Action.along` and it now mutates the caller and argument if they are ParallelActions, to save wasted instantiations and avoid GC.
+- **[UPDATE]** Added `/` operator to Action, which performs the non-mutating version of along, wrapping the caller and argument in a new ParallelAction.
+- **[CHANGE]** `Action.then` now mutates the caller and argument if they are SequenceActions, to save wasted instantiations and avoid GC. The `+` operator still performs the old non-mutating version, wrapping the caller and argument in a new SequenceAction.
 
 #### 1.9.10-b1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - **[CHANGE]** `Action.parallelTo` changed to `Action.along` and it now mutates the caller and argument if they are ParallelActions, to save wasted instantiations and avoid GC.
 - **[UPDATE]** Added `/` operator to Action, which performs the non-mutating version of along, wrapping the caller and argument in a new ParallelAction.
 - **[CHANGE]** `Action.then` now mutates the caller and argument if they are SequenceActions, to save wasted instantiations and avoid GC. The `+` operator still performs the old non-mutating version, wrapping the caller and argument in a new SequenceAction.
+- **[CHANGE]** `ParallelAction.plus()` and `SequenceAction.plus()` no longer unwrap their components.
 
 #### 1.9.10-b1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,14 @@
 #### 1.9.10-SNAPSHOT
-- **[CHANGE]** `Action.parallelTo` changed to `Action.along`.
-- **[CHANGE]** `Action.along` (formerly `Action.parallelTo`) and `Action.then` no longer unwrap the second action.
-- **[CHANGE]** `ParallelAction.along` (formerly `ParallelAction.parallelTo`) and `SequenceAction.then` simply add the second action to the group without unwrapping.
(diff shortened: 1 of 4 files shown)


cypherdare:
As discussed in #206, `plus` methods no longer unwrap.

actors/src/main/kotlin/ktx/actors/actions.kt
```diff
@@ -33,33 +33,51 @@ operator fun Actor.plusAssign(action: Action) = addAction(action)
...
- * @param action will be executed after this action.
+ * Wraps this action and the passed action in a [SequenceAction].
+ *
+ * @param action will be executed after [this] action.
  * @return [SequenceAction] storing both actions.
  */
-operator fun Action.plus(action: Action): SequenceAction = then(action)
+operator fun Action.plus(action: Action): SequenceAction = Actions.sequence(this, action)
+
+private fun SequenceAction.addUnwrapped(action: Action) {
```
czyzby: Shouldn't we remove this completely?

czyzby: Even `then` should not unwrap in my opinion, as this just leads to potential bugs.

CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
 #### 1.9.10-SNAPSHOT
+- **[CHANGE]** `Action.parallelTo` changed to `Action.along`.
+- **[CHANGE]** `Action.along` (formerly `Action.parallelTo`) and `Action.then` no longer unwrap the second action.
+- **[CHANGE]** `ParallelAction.along` (formerly `ParallelAction.parallelTo`) and `SequenceAction.then` simply add the second action to the group without unwrapping.
+- **[UPDATE]** Added `/` operator to Action, which performs the non-mutating version of along, wrapping the caller and argument in a new ParallelAction.
```
czyzby: `UPDATE` tag is reserved for dependencies' updates. New features such as this operator are marked with `FEATURE`.

actors/README.md
```diff
@@ -173,8 +174,7 @@ textField.onKeyboardFocusEvent { focusEvent, actor ->
 }
 ```
 
-Chaining actions with infix `then` function (`SequenceAction` utility):
-
+Chaining actions with infix `then` function (`SequenceAction` utility) and infix `along` function
```
czyzby: Since you haven't updated the example, revert this line. `along`/`/` usage examples would be welcome, although not necessary.

actors/README.md
```diff
@@ -44,10 +44,11 @@ be enough for most use cases.
...
 - Actions can be added and removed to individual `Actor` instances with `+=` and `-=` operators.
-- `Action.then` *infix* extension function allows to easily create action sequences with pleasant syntax.
-- `Action.parallelTo` *infix* extension function allows to easily create parallel actions with pleasant syntax.
-- `+` operator can be used to create action sequences (alternative syntax to `then`).
-- `+=` operator allows to add an action to an existing `SequenceAction`.
+- `Action.then` *infix* extension function allows easy creation of action sequences with pleasant syntax.
+- `Action.along` *infix* extension function allows easy creation of parallel actions with pleasant syntax.
+- `+` operator can be used to create action sequences (alternative non-mutating version of `then`).
+- `/` operator can be used combine actions in parallel (alternative non-mutating version of `along`).
+- `+=` operator allows to add an action to an existing `SequenceAction` or `ParallelAction`.
```
czyzby: If that's OK, you could add an additional section on when to use `then` vs `+` and `along` vs `/`.

Something along the lines of:
- Use operators for small amounts of actions (2-3) to prevent from creating too many objects. Prefer `then`/`along` for long chains of actions.
- Use operators to avoid mutating the sequences and parallel actions. If you are constructing actions in multiple steps and for any reason need a reference to intermediate step, use operators.
- Use `then`/`along` if you want to modify an existing sequence of parallel action without excessive creation of extra objects.

actors/src/main/kotlin/ktx/actors/actions.kt
```diff
@@ -33,65 +33,77 @@ operator fun Actor.plusAssign(action: Action) = addAction(action)
...
-  }
+/**
+ * Adds [action] to this [SequenceAction].
+ *
+ * @param action will be added to this [SequenceAction]
+ * @return [SequenceAction] this [SequenceAction]
+ */
+infix fun SequenceAction.then(action: Action): SequenceAction {
+  addAction(action)
+  return this
```
czyzby: Prefer `apply` blocks over returning `this`.

```kotlin
infix fun SequenceAction.then(action: Action): SequenceAction = apply { 
  addAction(action)
}
```

actors/src/main/kotlin/ktx/actors/actions.kt
```diff
@@ -33,65 +33,77 @@ operator fun Actor.plusAssign(action: Action) = addAction(action)
...
+/**
+ * Adds [action] to this [ParallelAction], so long as this isn't a [SequenceAction]. If it is a [SequenceAction],
+ * [@see Action.along].
+ *
+ * @param action will be added to this [ParallelAction].
+ * @return [ParallelAction] this [ParallelAction], now containing [action]]
+ */
+infix fun ParallelAction.along(action: Action): ParallelAction {
+  if (this is SequenceAction)
+    return this as Action along action // workaround for SequenceAction inheritance from ParallelAction
```
czyzby: I think this would be cleaner:

```kotlin
if (this is SequenceAction) {
  // Workaround for SequenceAction inheritance from ParallelAction:
  return Actions.parallel(this, action)
}
```

actors/src/main/kotlin/ktx/actors/actions.kt
```diff
@@ -33,65 +33,77 @@ operator fun Actor.plusAssign(action: Action) = addAction(action)
...
 
 /**
- * Wraps this action and the passed action with a [SequenceAction].
- * @param action will be executed after this action.
+ * Wraps this action and the passed action in a [SequenceAction].
+ *
+ * @param action will be executed after [this] action.
  * @return [SequenceAction] storing both actions.
  */
 operator fun Action.plus(action: Action): SequenceAction = then(action)
```
czyzby: Since `then` is now documented as mutating, let's not reuse it in a non-mutating operator and just reimplement it with `Actions.sequence` just to be future-proof.

actors/src/main/kotlin/ktx/actors/actions.kt
```diff
@@ -33,65 +33,77 @@ operator fun Actor.plusAssign(action: Action) = addAction(action)
...
+@Deprecated("parallelTo has been replaced with along.", ReplaceWith("this along action"), DeprecationLevel.WARNING)
+infix fun ParallelAction.parallelTo(action: Action): ParallelAction = along(action)
+
+/**
+ * Wraps this action and the passed action with a [ParallelAction].
+ *
+ * @param action will be executed at the same time as this action.
+ * @return [SequenceAction] storing both actions.
+ */
+operator fun Action.div(action: Action): ParallelAction = along(action)
```
czyzby: Since `along` is now documented as mutating, let's not reuse it in a non-mutating operator and just reimplement it with `Actions.parallel` just to be future-proof.

actors/README.md
cypherdare: I went ahead and added examples.

cypherdare: I added an example.

Issue #210, opened 2019-08-27
Enhance Batch.use


cypherdare:
Not sure if this would be getting too functionally specific, but it might be nice to set the projection matrix as we are calling `use` since that is done almost every time before calling `begin()`. Something like this:

```
/**
 * Automatically calls [Batch.begin] and [Batch.end].
 * @param projectionMatrix A projection matrix to set on the batch before [Batch.begin]. If null the batch's matrix
 * remains unchanged.
 * @param action inlined. Executed after [Batch.begin] and before [Batch.end].
 */
inline fun <B : Batch> B.use(projectionMatrix: Matrix4? = null, action: (B) -> Unit) {
  if (projectionMatrix != null)
    this.projectionMatrix = projectionMatrix
  begin()
  action(this)
  end()
}

/**
 * Automatically calls [Batch.begin] and [Batch.end].
 * @param camera The camera's [Camera.combined] matrix will be set to the batch's projection matrix before [Batch.begin]
 * @param action inlined. Executed after [Batch.begin] and before [Batch.end].
 */
fun <B : Batch> B.use(camera: Camera, action: (B) -> Unit) = use(camera.combined, action)
```

Thoughts?

czyzby:
It does seem like a common use case, good suggestion. Would you like to submit a PR?

czyzby:
Thanks for the contribution, snapshot release with the changes is being uploaded.

PR #211, opened 2019-08-30
Add matrix and camera parameters for batch.use

cypherdare:
An implementation of the discussion from #210.


graphics/src/test/kotlin/ktx/graphics/graphicsTest.kt
```diff
@@ -65,6 +67,21 @@ class GraphicsTest {
     val batch = mock<Batch>()
 
     batch.use {
+      verify(batch, never()).projectionMatrix = null // should not be set if none passed
```
czyzby: Use `any` matcher instead of null, as we want to make sure that the matrix is never changed.

czyzby:
Thanks for the PR, I'll push a snapshot release soon.

PR #216, opened 2019-10-02
Added more reified skin extension methods with default style name parameter

maltaisn:
Added more extension methods to Skin with reified parameters and/or default style name as default parameter value. Here's the changes in syntax with resources named "default":

```kotlin
// Optional
val res: Resource? = skin.optional("default", Resource::class.java)
val res: Resource? = skin.optional()

// Add
skin.add("default", resource)
skin.add(resource)
skin += resource

// Remove
skin.remove("default", Resource::class.java)
skin.remove<Resource>()

// Has
val found = skin.has("default", Resource::class.java)
val found = skin.has<Resource>()

// Get all
val map: ObjectMap<String, Resource>? = skin.getAll(Resource::class.java)
val map: ObjectMap<String, Resource>? = skin.getAll()
```

So now Skin can be fully used without having `Resource::class.java` type arguments and `"default"` name arguments. 

I feel like the latter is important since there is often only one resource of a type so it makes sense not to name it. LibGDX forces the use of `"default"` arguments which is ugly and more error-prone.

### DIFF
--- a/style/src/main/kotlin/ktx/style/style.kt
+++ b/style/src/main/kotlin/ktx/style/style.kt
@@ -21,6 +21,7 @@ import com.badlogic.gdx.scenes.scene2d.ui.Touchpad.TouchpadStyle
 import com.badlogic.gdx.scenes.scene2d.ui.Tree.TreeStyle
 import com.badlogic.gdx.scenes.scene2d.ui.Window.WindowStyle
 import com.badlogic.gdx.utils.GdxRuntimeException
+import com.badlogic.gdx.utils.ObjectMap
 import kotlin.annotation.AnnotationTarget.*
 
 /** Should annotate builder methods of Scene2D [Skin]. */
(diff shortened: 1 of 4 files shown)


style/src/main/kotlin/ktx/style/style.kt
```diff
@@ -84,8 +93,45 @@ inline operator fun <reified Resource : Any> Skin.set(name: String, resource: Re
...
-        this.set(name.toString(), resource)
+inline operator fun <reified Resource : Any, E : Enum<E>> Skin.set(name: E, resource: Resource) =
+    this.set(name.toString(), resource)
+
+/**
+ * Utility function that makes it easier to add [Skin] assets.
+ * @param name name of the passed resource. Defaults to [defaultStyle].
+ * @param resource will be added to the skin and mapped to the selected name.
+ */
+inline fun <reified Resource : Any> Skin.add(resource: Resource, name: String = defaultStyle) =
```
czyzby: Don't change the order of the parameters in relation to the original skin method.

maltaisn: I changed the order so the method can be called with the default name:
```kotlin
skin.add(resource)
```
Instead of
```kotlin
skin.add(resource = resource)
```
Sure someone could just use the `+=` method instead but personally I sometimes try to avoid overloaded operators.

czyzby: OK, makes sense.

maltaisn:
I updated the readme and changelog.

czyzby:
Thanks for the contribution!

czyzby:
Uploading snapshot with your changes.

maltaisn:
I just realized I missed the `Skin.addStyle` method that's still missing a default name value... do you think you can add it?

czyzby:
@maltaisn I'll look into it.

czyzby:
@maltaisn It's one of the older extension methods that is used internally - and probably in some applications of our users as well. I think the default style alternative is already covered by `add` extension method you added. I'd prefer to keep it as is to prevent API changes.

maltaisn:
Yes you're right. Turns out the new `add` extension is pretty well suited. Defining styles by code allows for immutable style classes so I found that there's really no need for builder style extensions on those.

PR #220, opened 2019-11-19
Direct reference Asset delegates

cypherdare:
Untested--submitting for discussion.

Not sure if this is too invasive a change. The idea here is to avoid AssetManager having to do the map look-up every time we access our Texture (or whatever) property when using delegates for asset properties.

By the way...any idea my ktx forked project in IntelliJ might no longer recognize junit and ockito libraries? I haven't opened the ktx project in a few months, but it was fine before.


cypherdare:
Sorry, wrong branch.

PR #221, opened 2019-11-19
Add asset reference caching to ManagedAsset

cypherdare:
Untested--submitting for discussion.

Not sure if this is too invasive a change. The idea here is to avoid AssetManager having to do the map look-up every time we access our Texture (or whatever) property when using delegates for asset properties.

By the way...any idea my ktx forked project in IntelliJ might no longer recognize junit and ockito libraries? I haven't opened the ktx project in a few months, but it was fine before.


czyzby:
Getting a loaded asset from the manager should be relatively cheap, even if the method is `synchronized`. Introducing cache might unfortunately bring some issues, such as stale assets if they happen to be unloaded or replaced directly within the manager. Unless you could find a way of checking if the cache is up to date - with it being cheaper than just getting the asset from the manager - I don't we should introduce this change.

> By the way...any idea my ktx forked project in IntelliJ might no longer recognize junit and ockito libraries? I haven't opened the ktx project in a few months, but it was fine before.

I haven't had this problem, but you could try refreshing Gradle or invalidating caches.

Issue #222, opened 2019-11-23
AssetContainer


cypherdare:
This is an idea for utilizing the asset delegate classes to easily manage groups of assets with minimal typing. I used a similar concept for CoveTools' [AssignmentAssetManager](https://github.com/CypherCove/CoveTools#assignmentassetmanager), but the usage of that is ugly in Kotlin due to null-safety, and I think this could be more feature rich.

Usage would be to subclass AssetContainer to define groups of assets by delegate so they could easily be loaded and unloaded. Here are some examples.

**Complex game with groups of assets for different screens:**
```kotlin
// Define some containers like this:
class PauseScreenAssets: AssetContainer(
        assetManager, 
        pathPrefix = "pause/"){
    val skin by delayedAsset<Skin>("mySkin.json")
    val music by delayedAsset<Music>("pauseMusic.ogg")
    //...
}

// Load and unload whole containers:
fun goToMapScreen(){
    stage1Assets.unloadAll()
    with(mapAssets) {
        loadAll()
        finishLoading()
    }
    //...
}
```

**Simple game where all assets are loaded at once:**
We can almost do this already, but it's a little easier to read through the list of assets when they don't all have `assetManager.load` written before them.
All the assets can be defined in a single file, and a default built-in AssetManager can be used so the whole thing is wrapped to a simpler object that lists all the Textures, Sounds, Shaders, etc.
```kotlin
class Assets: AssetsContainer() {
    val atlas by asset<TextureAtlas>("myAtlas")
    val logo by asset<Texture>("logo.png")
    //...

    override fun onLoadingFinished() {
        // Here's a hook for acquiring AtlasRegion references, etc.
    }
}
```

**Work with specific Asset objects even though they were used as delegates (requires reflection):**
```kotlin
// Within the AssetContainer subclass:
::myTexture.toAsset().unload()

// Outside the class:
myAssetContainer.getAsset{::myTexture}.unload()
```

Here's how I've outlined a prototype. If this kind of detailed helper class is a good fit for KTX, we can discuss improving on this and making it more elegant. (Not sure if it's in KTX's scope to go this far with a helper class.)

```kotlin
open class AssetContainer(val assetManager: AssetManager = AssetManager(),
                          val filePrefix: String = ""): Disposable {
    private val assets = com.badlogic.gdx.utils.Array<Asset<*>>()

    fun add(asset: Asset<*>) = assets.add(asset)

    fun remove(asset: Asset<*>) = assets.removeValue(asset, true)

[...]

PR #223, opened 2019-11-27
AssetGroup

cypherdare:
This is for #222 .

A few notes:

- `members` is protected so it can be accessed by the protected `asset()` and `delayedAsset()` inline functions. There's also an internal `@PublishedApi` for it so it can be used for the `unloadAllSafely()` inline function. But I thought in some cases someone might want to customize the `asset()`/`delayedAsset()` functions and therefore need a way to directly add to the set.
- I added `toString()` for the two types of assets so reacting to caught exceptions in `unloadAllSafely()`  gives you something that is useful to log. Otherwise, you have very little clue as to which asset threw the exception.
- I made MockAsset throw an exception if disposed twice, which is something some Gdx classes do, so I thought it would help effectively test safe unloading.

### DIFF
--- a/assets/README.md
+++ b/assets/README.md
@@ -63,16 +63,28 @@ such as `loadAll()` or `unloadAll()`. The intended use is to subclass `AssetGrou
 list its member assets as properties using `AssetGroup.asset()` or `AssetGroup.delayedAsset()`. It also allows for using 
 a common prefix for the file names of the group in case they are stored in a specific subdirectory. For example:
 ```Kotlin
-class MapScreenAssets(manager: AssetManager) : AssetGroup(manager, "mapScreen/"){
+/** A group of assets that are stored in the mapScreen directory and are used only on the map screen. */
+class MapScreenAssets(manager: AssetManager) : AssetGroup(manager, filePrefix = "mapScreen/"){
     val atlas by asset<TextureAtlas>("atlas.json")
     val music by asset<Music>("mapScreen.ogg")
 }
 
 val mapScreenAssets = MapScreenAssets(manager)
-//...
-manager.finishLoading()
-//...
-mapScreenAssets.dispose()
+
+// Then, when ready to load them:
+// Using asset() to create the assets pre-queues them once the group is instantiated.
+// If you were using delayedAsset() instead, you would call mapScreenAssets.loadAll() when ready to queue them.
+if (mapScreenAssets.update()) {
+    // The member assets are now ready to use.
+} else {
+    // Continue showing loading screen, for example.
+}
+
+// Alternatively, mapScreenAsset.finishLoading() would block until the assets are loaded, possibly finishing earlier 
+// than manager.finishLoading() would.
+
+// When finished with these assets:
+mapScreenAssets.unloadAll()
 ```
 
 Note: if you can use coroutines in your project, [`ktx-assets-async`](../assets-async) module provides a lightweight
--- a/assets/README.md
+++ b/assets/README.md
@@ -58,6 +58,22 @@ loaded in the first place. Typical usage: `assetManager.unloadSafely("test.png")
 exception thrown during reloading. Note that `AssetManager` can throw `GdxRuntimeException` if the asset was not loaded yet.
 - `AssetManager.getLoader` and `setLoader` extension methods with reified types added to ease handling of `AssetLoader`
 instances registered in the `AssetManager`.
+- The `AssetGroup` class is provided for easily grouping together assets so they can be managed as a group through calls 
+such as `loadAll()` or `unloadAll()`. The intended use is to subclass `AssetGroup` and 
+list its member assets as properties using `AssetGroup.asset()` or `AssetGroup.delayedAsset()`. It also allows for using 
+a common prefix for the file names of the group in case they are stored in a specific subdirectory. For example:
+```Kotlin
+class MapScreenAssets(manager: AssetManager) : AssetGroup(manager, "mapScreen/"){
+    val atlas by asset<TextureAtlas>("atlas.json")
+    val music by asset<Music>("mapScreen.ogg")
+}
+
+val mapScreenAssets = MapScreenAssets(manager)
+//...
+manager.finishLoading()
+//...
+mapScreenAssets.dispose()
+```
 
 Note: if you can use coroutines in your project, [`ktx-assets-async`](../assets-async) module provides a lightweight
 coroutines-based alternative to `AssetManager` that can greatly simplify your asset loading code.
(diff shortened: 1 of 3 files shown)


assets/README.md
```diff
@@ -58,6 +58,22 @@ loaded in the first place. Typical usage: `assetManager.unloadSafely("test.png")
 exception thrown during reloading. Note that `AssetManager` can throw `GdxRuntimeException` if the asset was not loaded yet.
 - `AssetManager.getLoader` and `setLoader` extension methods with reified types added to ease handling of `AssetLoader`
 instances registered in the `AssetManager`.
+- The `AssetGroup` class is provided for easily grouping together assets so they can be managed as a group through calls 
+such as `loadAll()` or `unloadAll()`. The intended use is to subclass `AssetGroup` and 
+list its member assets as properties using `AssetGroup.asset()` or `AssetGroup.delayedAsset()`. It also allows for using 
+a common prefix for the file names of the group in case they are stored in a specific subdirectory. For example:
+```Kotlin
+class MapScreenAssets(manager: AssetManager) : AssetGroup(manager, "mapScreen/"){
```
czyzby: Add named parameter before the `"mapScreen/"` to clarify the example.

assets/README.md
```diff
@@ -58,6 +58,22 @@ loaded in the first place. Typical usage: `assetManager.unloadSafely("test.png")
...
+    val atlas by asset<TextureAtlas>("atlas.json")
+    val music by asset<Music>("mapScreen.ogg")
+}
+
+val mapScreenAssets = MapScreenAssets(manager)
+//...
```
czyzby: Replace `...` comments with descriptions of the section. What is the user supposed to do here? `loadAll`?

assets/README.md
```diff
@@ -58,6 +58,22 @@ loaded in the first place. Typical usage: `assetManager.unloadSafely("test.png")
...
+manager.finishLoading()
```
czyzby: README examples should generally showcase the most common use case. I'd argue that `finishLoading` would be rarely used in practice. Could you replace it with `loadAll`/`update` loop or something similar?

czyzby: Two separate examples for `delayedAsset`/`asset` showcasing would also be nice.

assets/src/main/kotlin/ktx/assets/assets.kt
```diff
@@ -212,3 +215,99 @@ inline fun <reified Type : Any, Parameters : AssetLoaderParameters<Type>> AssetM
...
+      member.unload()
+  }
+
+  /** Unloads all of the assets in this group, catching any exceptions. If any of the assets are dependencies of
+   * assets outside this group, or if they were loaded more than once, their load counts will only be decremented by one
+   * and the associated [AssetManager] will retain them.
+   * @param onError Called in response to each caught exception. By default, the exceptions are ignored.
+   * */
+  inline fun unloadAllSafely(onError: (Asset<*>, Exception) -> Unit = { _, exception -> exception.ignore() }) {
+    for (member in `access$members`){
```
czyzby: That's an unusual syntax. For the sake of clarity, this function does not have to be inlined - it's unlikely to be performance critical either way.

assets/src/main/kotlin/ktx/assets/assets.kt
```diff
@@ -212,3 +215,99 @@ inline fun <reified Type : Any, Parameters : AssetLoaderParameters<Type>> AssetM
...
+   * @param params optional asset loading parameters which might affect how the assets are loaded. Can be null.
+   * @return an [Asset], registered as a member of this AssetGroup.
+   * */
+  protected inline fun <reified T : Any> delayedAsset(fileName: String, params: AssetLoaderParameters<T>? = null) =
+      manager.loadOnDemand("$filePrefix$fileName", params).also { members.add(it) }
+
+  @PublishedApi
+  internal val `access$members`: ObjectSet<Asset<*>>
+    get() = members
+}
```
czyzby: Missing new line.

assets/src/test/kotlin/ktx/assets/assetsTest.kt
```diff
@@ -477,4 +559,5 @@ class MockAssetLoader(fileHandleResolver: FileHandleResolver) :
 
   /** Allows to set [MockAsset.additional] via loader. Tests assets parameters API. */
   class MockParameter(val additional: String?) : AssetLoaderParameters<MockAsset>()
-}
+
+}
```
czyzby: Missing new line.

assets/src/main/kotlin/ktx/assets/assets.kt
cypherdare: I was inlining this more so to allow the response to an error be to return early. But I suppose that would be a very rare case. You would either be using this to debug, or to unload everything regardless of error.

assets/README.md
cypherdare: I left a comment explaining how you'd use it differently. Keep it concise, or add a second example class to expand on the differences?

czyzby: I think 2 separate examples is the way to go, since there seem to be 2 approaches to managing assets with groups and joining can be tricky. You can move them to the `Usage examples` section.

assets/src/main/kotlin/ktx/assets/assets.kt
czyzby: Not sure if you'd like to or should break early here, to be honest.

cypherdare: I suppose if someone wanted to break early, they would use wrap an `unloadAll()` call in a try/catch instead of using `unloadAllSafely()`.

assets/README.md
cypherdare: I expanded it in a concise way. Seems OK to keep in this section.

cypherdare:
It just occurred to me, `finishLoading()` possibly could prioritize the group's members if we hung onto the asset file names and used `manager.finishLoadingAsset<Any>(filename)` on each of them. We wouldn't be using the return value of that method. This could be done by making `members` into an `ObjectMap<Asset<*>, String>` or by promoting `assetDescriptor` to part of the Asset interface.

czyzby:
> It just occurred to me, `finishLoading()` possibly could prioritize the group's members if we hung onto the asset file names and used `manager.finishLoadingAsset<Any>(filename)` on each of them.

That's only partially true, because AFAIR it does not affect loading scheduling at all - you just happen to break early if the asset was loaded sooner than later. I think we could leverage that, since asset descriptors already store the names and could possibly be parts of the `Asset` APIs.

cypherdare:
OK, I implemented that change. it looks like a cleaner way to do it, because we don't have to repeatedly check `isLoaded()` for the whole set of assets over and over

It's technically breaking for anybody that wrote their own Asset implementation, although I can't think of a reason anyone would have done that.

czyzby:
Yeah, it's unlikely to be an issue.

czyzby:
Thanks for the contribution.

Issue #226, opened 2019-12-04
In-place functions for GdxArray


cypherdare:
It would  be nice to have `removeAll(predicate)` (filter in place), and possibly `retainAll` (although it's a bit redundant).

Side question: why doesn't Pooledlist implement MutableIterable? It would be nice to have all those stlib extension functions available.

czyzby:
That's a good suggestion, would you like to submit a PR?

> Side question: why doesn't Pooledlist implement MutableIterable? It would be nice to have all those stlib extension functions available.

Probably because LibGDX don't implement any stdlib collections interfaces either, but I don't see why we couldn't add this.

cypherdare:
Yes, I'll take it on. I can try making PooledList a MutableIterable as well. I don't know the reasoning behind the Gdx classes not implementing Collection.

czyzby:
Me neither, and honestly it's been tedious to write any extensions for them, as they do not share any custom interfaces either.

czyzby:
Thanks for another great contribution @cypherdare, the new utilities are currently being uploaded via snapshot releases.

By the way, I'd like to publish another beta release soon, but if you've got any more ideas or plan more contributions, let me know. We can wait for #227.

cypherdare:
I'm not entirely sure I'll get to #277 before January, so don't wait if you want to get it out sooner than that.

czyzby:
That's fine, take your time.

PR #228, opened 2019-12-10
removeAll and retainAll for GdxArray

cypherdare:
For #226.

### DIFF
--- a/collections/README.md
+++ b/collections/README.md
@@ -32,7 +32,8 @@ chained.
 - Missing `addAll` and `removeAll` methods for arrays and iterables were added.
 - `iterate` method allows to iterate over collection's elements, while providing reference to `MutableInterator`. Can be
 used to easily remove collection elements during iteration.
-- `map`, `filter`, `flatten` and `flatMap` methods that work like methods in Kotlin stdlib but return `GdxArray`.
+- `removeAll` and `retainAll` higher order functions that work like functions in Kotlin stdlib
+- `map`, `filter`, `flatten` and `flatMap` functions that work like functions in Kotlin stdlib but return `GdxArray`.
 - Every iterable and array can be converted to `Array` using `toGdxArray` method.
 - `IntArray`, `BooleanArray` and `FloatArray` can be converted to corresponding LibGDX primitive collections using
 `toGdxArray` method.
--- a/collections/README.md
+++ b/collections/README.md
@@ -32,8 +32,8 @@ chained.
 - Missing `addAll` and `removeAll` methods for arrays and iterables were added.
 - `iterate` method allows to iterate over collection's elements, while providing reference to `MutableInterator`. Can be
 used to easily remove collection elements during iteration.
-- `removeAll` and `retainAll` higher order functions that work like functions in Kotlin stdlib
-- `map`, `filter`, `flatten` and `flatMap` functions that work like functions in Kotlin stdlib but return `GdxArray`.
+- `removeAll` and `retainAll` higher-order functions that work like functions in Kotlin stdlib
+- `map`, `filter`, `flatten` and `flatMap` methods that work like methods in Kotlin stdlib but return `GdxArray`.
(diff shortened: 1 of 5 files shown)


collections/src/test/kotlin/ktx/collections/arraysTest.kt
```diff
@@ -364,6 +364,25 @@ class ArraysTest {
...
+
+  @Test
+  fun `should retain elements from existing GdxArray`() {
+    val array = GdxArray.with(1, 2, 3, 4, 5)
+
+    array.retainAll { it % 2 == 1 }
+
+    assertEquals(GdxArray.with(1, 3, 5), array)
+  }
+
```
czyzby: Add tests for edge cases: empty arrays, no elements retained/removed.

collections/README.md
```diff
@@ -32,7 +32,8 @@ chained.
 - Missing `addAll` and `removeAll` methods for arrays and iterables were added.
 - `iterate` method allows to iterate over collection's elements, while providing reference to `MutableInterator`. Can be
 used to easily remove collection elements during iteration.
-- `map`, `filter`, `flatten` and `flatMap` methods that work like methods in Kotlin stdlib but return `GdxArray`.
+- `removeAll` and `retainAll` higher order functions that work like functions in Kotlin stdlib
```
czyzby: Could you also update the `CHANGELOG.md` file?

collections/src/main/kotlin/ktx/collections/arrays.kt
```diff
@@ -225,6 +225,40 @@ inline fun <Type, R : Comparable<R>> GdxArray<out Type>.sortByDescending(crossin
   if (size > 1) this.sort(compareByDescending(selector))
 }
 
+/**
+ * Removes elements from the array that satisfy the predicate.
+ */
+inline fun <Type> GdxArray<Type>.removeAll(predicate: (Type) -> Boolean) {
+  var currentWriteIndex = 0
+  for (i in 0 until size){
```
czyzby: I don't think this file was formatted with KTX formatter.

cypherdare:
I changes `this[]` to `items[]` to skip the unnecessary bounds check, since Array already provides direct access to the backing array for this sort of thing. Kind of trivial, but might as well.

What would you think about making overloads of these functions as well as `filter` that let you pass a pool to return removed items to? I could see myself using that a lot. Could be a separate PR.

collections/src/test/kotlin/ktx/collections/arraysTest.kt
cypherdare: Also added cases for all elements retained/removed.

czyzby:
> What would you think about making overloads of these functions as well as filter that let you pass a pool to return removed items to? I could see myself using that a lot. Could be a separate PR.

Seems like this could be simply returned by the functions, but I get how this is pretty situational and an extra parameter would be more efficient and customizable. I think you should submit a separate PR for that.

cypherdare:
> > What would you think about making overloads of these functions as well as filter that let you pass a pool to return removed items to? I could see myself using that a lot. Could be a separate PR.
> 
> Seems like this could be simply returned by the functions, but I get how this is pretty situational and an extra parameter would be more efficient and customizable. I think you should submit a separate PR for that.

The idea arose as I realized I'd rarely be using these methods for what I'm usually doing with in-place removal, which is tracking many game objects that expire on a regular basis, usually particles. So the only thing I ever do with expired objects is either lose them to the GC or free them to a pool.

If we return all removed objects, we are allocating an array, which halfway defeats the purpose of removing in-place, and in many cases users would be ignoring the returned values anyway.

If we provide a parameter for a second lambda that is a handler for removed items, the syntax gets a little clunky for something that ~probably~ would only be used to return items to a pool (my opinion/guess).

So my vote is for just a pool. I'm questioning whether it should be a nullable optional parameter, or full overloads of the functions should be written, since they're inline and already kind of long.

PR #230, opened 2019-12-17
Update rough line count of the file

Stuie:
The documentation says the file has around 150 lines including documentation, but it's closer to 230.

Might just make sense to remove the number of lines altogether to avoid having to update it ever again.

### DIFF
--- a/inject/README.md
+++ b/inject/README.md
@@ -17,7 +17,7 @@ its inline functions, allows to omit the reflection and annotations usage altoge
 DSL.
 
 Why not use an existing Kotlin DI library? `ktx-inject` is a tiny extension consisting a single source file with about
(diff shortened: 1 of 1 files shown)


czyzby:
I noticed that as well. :) If you check out the [`develop`](https://github.com/libktx/ktx/tree/develop/inject) branch, which contains the latest changes, you'll see that this particular line was changed to "consisting a single source file with **a few hundred** lines".

I know that having the `master` branch without the latest changes as the default one might a bit confusing, but then again - I wanted the repository to represent the version that the users are likely importing into their project, so they can easily go through the sources and up-to-date documentation without unreleased features.

Thanks for the pull request, but since it's already addressed, I'm going to close it. :)

czyzby:
I also noticed there's a small error in this paragraph. Thanks again.

PR #232, opened 2019-12-20
Optional pool freeing when using removeAll or retainAll

cypherdare:
These two functions could not be used if you want to free the removed items to a pool. The alternative would be to return the removed items, but that would create allocation/GC churn for something that is not always needed. And by far the most likely thing you would ever do with a removed item, if anything,  is return it to a pool.

I elected not to add this to `filter` as you would likely not use `filter` with pooled items. Pooled items are used in situations where the array is probably large and allocation/GC churn is trying to be avoided, so you would likely also want to be doing an in-place filter with `retainAll`.

### DIFF
--- a/collections/README.md
+++ b/collections/README.md
@@ -32,7 +32,8 @@ chained.
 - Missing `addAll` and `removeAll` methods for arrays and iterables were added.
 - `iterate` method allows to iterate over collection's elements, while providing reference to `MutableInterator`. Can be
 used to easily remove collection elements during iteration.
-- `removeAll` and `retainAll` higher-order functions that work like functions in Kotlin stdlib
+- `removeAll` and `retainAll` higher-order functions that work like functions in Kotlin stdlib. A Pool can optionally be 
+passed to automatically free the removed items.
 - `map`, `filter`, `flatten` and `flatMap` methods that work like methods in Kotlin stdlib but return `GdxArray`.
 - Every iterable and array can be converted to `Array` using `toGdxArray` method.
 - `IntArray`, `BooleanArray` and `FloatArray` can be converted to corresponding LibGDX primitive collections using
--- a/collections/src/main/kotlin/ktx/collections/arrays.kt
+++ b/collections/src/main/kotlin/ktx/collections/arrays.kt
@@ -6,16 +6,22 @@ import com.badlogic.gdx.utils.Pool
 
 /** Alias for [com.badlogic.gdx.utils.Array] avoiding name collision with the standard library. */
 typealias GdxArray<Element> = com.badlogic.gdx.utils.Array<Element>
+
 /** Alias for [com.badlogic.gdx.utils.BooleanArray] avoiding name collision with the standard library. */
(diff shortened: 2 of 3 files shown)


collections/src/main/kotlin/ktx/collections/arrays.kt
```diff
@@ -233,8 +235,9 @@ inline fun <Type, R : Comparable<R>> GdxArray<out Type>.sortByDescending(crossin
 
 /**
  * Removes elements from the array that satisfy the [predicate].
+ * @param pool Removed items are freed to this pool.
  */
-inline fun <Type> GdxArray<Type>.removeAll(predicate: (Type) -> Boolean) {
+inline fun <Type> GdxArray<Type>.removeAll(pool: Pool<Type>?, predicate: (Type) -> Boolean) {
```
czyzby: You should add a default `null` value for `pool` parameter.

collections/README.md
```diff
@@ -32,7 +32,8 @@ chained.
 - Missing `addAll` and `removeAll` methods for arrays and iterables were added.
 - `iterate` method allows to iterate over collection's elements, while providing reference to `MutableInterator`. Can be
 used to easily remove collection elements during iteration.
-- `removeAll` and `retainAll` higher-order functions that work like collection extensions in Kotlin stdlib.
+- `removeAll` and `retainAll` higher-order functions that work like collection extensions  in Kotlin stdlib. A Pool can
```
czyzby: Remove double space.

czyzby:
Using `Pool` is a pretty specific use case. I think returning an array of the removed elements would be more flexible - although less performant and excessive in many cases. Consuming two lambdas, one of which would be used on the removed elements, is also an option, but then again - it makes the syntax more convoluted. Maybe we should add another method such as `selectAll` or something similar that would return the removed elements? I'm not sure.

cypherdare:
Maybe this new method could take an optional array to put the removed
elements in? If null is passed, one could be allocated. This way there is
the option to use a reusable array. The function could return the other
array to allow immediate actions to be done on the removed elements.

czyzby:
Also thought about this, since LibGDX seems to use this pattern sometimes. I think we should use this over the pool. Would you like to apply this change?

cypherdare:
You mean add this function and remove the pool-freeing option?

And should it maybe be called something like “transfer” since “selectAll”
might sound ambiguous about whether the items are removed?

I would personally like an option to at least pass a removal handler lambda
to the other two methods if we’re removing the pool option.

czyzby:
`transfer` sounds good, we can leave the `Pool` utilities and add `transfer` that consumes an `Array` instance and a lambda. Let me know if you want to add `transfer` in this PR or if you'd like to do this separately.

cypherdare:
OK. I'll add it to this one. Could be a few days.

czyzby:
No problem, thanks again.

czyzby:
I'll go ahead and finish the PR, I hope you didn't start working on it yet. Thanks for the contribution!

czyzby:
Done, you can view the changes in the latest linked commit.

Issue #233, opened 2019-12-20
Tiled extensions


Quillraven:
Hi,

just wanted to know if there is an interest in TiledMap extensions as well.
I am using a few in my own game and I could make a pull request if someone is interested.

Just needed to know if this is something useful or not and if it is useful, what do I need to do for a valid pull request? :)

I have following extensions in place for accessing properties
```kotlin
// extension method to access properties the Kotlin way ;)
inline fun <reified T> MapObject.property(key: String, defaultValue: T): T =
    this.properties[key, defaultValue, T::class.java]

inline fun <reified T> MapLayer.property(key: String, defaultValue: T): T =
    this.properties[key, defaultValue, T::class.java]
```
This extension is used to retrieve the shape of a MapObject
```kotlin
// extension property to access shape of a MapObject
val MapObject.shape: Shape2D
    get() = when (this) {
        is RectangleMapObject -> this.rectangle
        is PolylineMapObject -> this.polyline
        is PolygonMapObject -> this.polygon
        else -> {
            LOG.error { "Unsupported MapObject of type ${this::class.java}. Cannot retrieve shape!" }
            Rectangle.tmp.set(0f, 0f, 1f, 1f)
        }
    }
```

Extensions to quickly access properties that are used very often like id, x, y, width, height
```kotlin
// extension property to easily access MapObject IDs, x and y position
val MapObject.id: Int
    get() = this.property(PROPERTY_ID, -1)
val MapObject.x: Float
    get() = this.property(PROPERTY_X, 0f)
val MapObject.y: Float
    get() = this.property(PROPERTY_Y, 0f)

// inside my own Map class, which stores a reference to  a TiledMap I also have
val width: Float
        get() = property(PROPERTY_WIDTH, 0f) * property(PROPERTY_TILE_WIDTH, 0f) * UNIT_SCALE
    val height: Float
        get() = property(PROPERTY_HEIGHT, 0f) * property(PROPERTY_TILE_HEIGHT, 0f) * UNIT_SCALE

 // extension method to access properties the Kotlin way for TiledMap
    inline fun <reified T> property(key: String, defaultValue: T): T =
        tiledMap.properties[key, defaultValue, T::class.java]

    fun containsProperty(key: String) = tiledMap.properties.containsKey(key)

    private fun layer(layerName: String): MapLayer {
        val layer = tiledMap.layers.get(layerName)
        if (layer == null) {
            // layer not defined in tiled map
[...]

PR #234, opened 2019-12-21
#233 tile map extensions for libktx

Quillraven:
First draft of the extensions that I am using in my projects. I had to generalize them a little bit and also tried to follow the kotlin "standard" with naming of some functions like e.g. "propertyOrNull".

Let me know what you think :)

Looking forward to your feedback!


tiled/README.md
```diff
@@ -0,0 +1,127 @@
...
+- `totalWidth()` and `totalHeight()`
+- `layer(layerName)`
+- `forEachMapObject(layerName, action)`
+
+### Usage examples
+
+`MapObject`:
+
+```kotlin
+// assume mapObj refers to a MapObject instance
```
czyzby: You can make it explicit without an object by starting the example with e.g. `val mapObj: MapObject = getMapObject()`.

tiled/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+// before
+val myProp = mapObj.properties.get("myProperty", "", String::class.java)
+val myOtherProp = mapObj.properties.get("myOtherProperty", Int::class.java)
+val x = mapObj.properties.get("x", 0f, Float::class.java)
+val y = mapObj.properties.get("y", 0f, Float::class.java)
+val shape = (mapObj as RectangleMapObject).rectangle
```
czyzby: We generally tend to avoid `before`/`after` showcases in the examples section, and instead include in the guide (if necessary). Usage examples should be limited to idiomatic KTX code that you can base your application on.

tiled/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+// using libktx
+val myProp = mapObj.property("myProperty", "")
```
czyzby: Explicit type might help here.

tiled/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+val myLayerProp = map.layers.get("myLayer")?.properties?.get("myLayerProp", Int::class.java)
+
+
+
+// using libktx
+val myProp:String? = map.propertyOrNull("myProp")
+val myProp2 = map.property("myProp2", true)
+val width = map.width
+val totalWidth = map.totalWidth()
```
czyzby: A lot of these properties seem loosely connected, especially to the iteration example. These could be separate examples with some short introduction, e.g.:

--- 

Iterating over all map objects of a `TiledMap` in a selected layer:

```kotlin
map.forEachMapObject("myLayer") {
    val x = it.x
    val y = it.y
    // create an object at x/y position
}
```

tiled/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+map.forEachMapObject("myLayer") {
+    val x = it.x
+    val y = it.y
+    // create an object at x/y position
+}
+
+val myLayerProp: Int? = map.layer("myLayer").propertyOrNull("myLayerProp")
+```
+
```
czyzby: Please include necessary imports in the examples. E.g. `import ktx.tiled.*`. The examples should work (or be easy to fix) after copy-pasting the code.

tiled/gradle.properties
```diff
@@ -0,0 +1,2 @@
+projectName=ktx-tiled
+projectDesc=TiledMap utilities for LibGDX applications written with Kotlin.
```
czyzby: Maybe just `Tiled utilities`?

tiled/src/main/kotlin/ktx/tiled/mapLayer.kt
```diff
@@ -0,0 +1,28 @@
...
+ * @return value of the property or null if the property is missing
+ */
+inline fun <reified T> MapLayer.propertyOrNull(key: String): T? = this.properties[key, T::class.java]
+
+/**
+ * Extension method to directly access the [MapProperties] of a [MapLayer] and its [containsKey][MapProperties.containsKey] method
+ * @param key property name
+ * @return true if the property exists. Otherwise false
+ */
+fun MapLayer.containsProperty(key: String) = properties.containsKey(key)
```
czyzby: Missing new line at the end of the file.

czyzby: This function could be the contains operator, so that `in` would work with a `MapLayer`. Not sure if we want to use it for properties though.

tiled/src/main/kotlin/ktx/tiled/mapObject.kt
```diff
@@ -0,0 +1,73 @@
...
+import com.badlogic.gdx.math.*
+
+/**
+ * Extension method to directly access the [MapProperties] of a [MapObject]. The type is automatically
+ * derived from the type of the given default value. If the property is not defined the defaultValue will be returned.
+ * @param key property name
+ * @param defaultValue default value in case the property is missing
+ * @return value of the property or defaultValue if property is missing
+ */
+inline fun <reified T> MapObject.property(key: String, defaultValue: T): T = this.properties[key, defaultValue, T::class.java]
```
czyzby: Do `MapObject`, `MapLayer` etc. follow any sort of class hierarchy? Maybe the utilities do not have to be added 1 by 1.

tiled/src/main/kotlin/ktx/tiled/mapObject.kt
```diff
@@ -0,0 +1,73 @@
...
+ * @return true if the property exists. Otherwise false
+ */
+fun MapObject.containsProperty(key: String) = properties.containsKey(key)
+
+/**
+ * Extension property to retrieve the x-coordinate of the [MapObject]. If the object does not have
+ * a x-coordinate then 0 is returned
+ */
+val MapObject.x: Float
+    get() = property("x", 0f)
```
czyzby: Is this property always defined? An unexpected 0 without any errors or warnings could be misleading. I'd consider converting it to `propertyOrNull("x") ?: throw MissingPropertyException("x")`.

Now that I think about it, `property(name: String)` extension method that return a non-null value or throws an exception would also be a useful utility.

tiled/src/main/kotlin/ktx/tiled/mapObject.kt
```diff
@@ -0,0 +1,73 @@
...
+ * Note that [TextureMapObject] is not supported by this method as it has no related shape.
+ */
+val MapObject.shape: Shape2D
+    get() = when (this) {
+        is CircleMapObject -> this.circle
+        is EllipseMapObject -> this.ellipse
+        is PolylineMapObject -> this.polyline
+        is PolygonMapObject -> this.polygon
+        is RectangleMapObject -> this.rectangle
+        else -> throw UnsupportedOperationException("Shape extension function is not supported for MapObject of type ${this::class.java}")
```
czyzby: This exception might be a bit too generic, I'd add something along `MissingPropertyException` in `ktx-tiled` module to handle this and similar cases.

tiled/src/main/kotlin/ktx/tiled/tiledMap.kt
```diff
@@ -0,0 +1,99 @@
+package ktx.tiled
+
+import com.badlogic.gdx.maps.MapLayer
+import com.badlogic.gdx.maps.MapObject
+import com.badlogic.gdx.maps.MapProperties
+import com.badlogic.gdx.maps.tiled.TiledMap
+
+private val DEFAULT_MAP_LAYER = MapLayer().apply { name = "ktx-default-map-layer" }
```
czyzby: I don't think this is a good approach, especially since `MapLayer` objects seem to be mutable.

tiled/src/main/kotlin/ktx/tiled/tiledMap.kt
```diff
@@ -0,0 +1,99 @@
...
+ * @return value of the property or null if the property is missing
+ */
+inline fun <reified T> TiledMap.propertyOrNull(key: String): T? = this.properties[key, T::class.java]
+
+/**
+ * Extension method to directly access the [MapProperties] of a [TiledMap] and its [containsKey][MapProperties.containsKey] method
+ * @param key property name
+ * @return true if the property exists. Otherwise false
+ */
+fun TiledMap.containsProperty(key: String) = properties.containsKey(key)
```
czyzby: This could also be an operator.

tiled/src/main/kotlin/ktx/tiled/tiledMap.kt
```diff
@@ -0,0 +1,99 @@
...
+ */
+val TiledMap.tileWidth: Int
+    get() = property("tilewidth", 0)
+
+/**
+ * Extension property to retrieve the tile height of each tile of the [TiledMap]. If the map does not have
+ * a **tileheight** property then 0 is returned
+ */
+val TiledMap.tileHeight: Int
+    get() = property("tileheight", 0)
```
czyzby: Same comment about defaults applies here. I think I'd rather have to handle an exception rather than get an unexpected 0 on tile height, width, etc.

tiled/src/main/kotlin/ktx/tiled/tiledMap.kt
czyzby: Maybe a `contains` operator would work better to check for existing layers in case of a `TiledMap` though.

tiled/src/main/kotlin/ktx/tiled/tiledMap.kt
```diff
@@ -0,0 +1,99 @@
...
+ */
+fun TiledMap.totalHeight() = height * tileHeight
+
+/**
+ * Extension method to retrieve a non-null [MapLayer] of the [TiledMap]. In case the layer
+ * cannot be found, a default empty layer is returned with name **ktx-default-map-layer**.
+ * @param layerName name of [MapLayer]
+ * @return [MapLayer] of given name or an empty default map layer, if the specified layer does not exist
+ */
+fun TiledMap.layer(layerName: String) = layers[layerName] ?: DEFAULT_MAP_LAYER
```
czyzby: OK, getting a default instance of a `MapLayer` would not be very useful and it could lead to unexpected behaviors. I wouldn't have suspected it as a user. You should definitely throw an exception if the map layer is missing.

tiled/src/main/kotlin/ktx/tiled/tiledMap.kt
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Extension method to easily execute an action per [MapObject] of a given [MapLayer].
+ * @param layerName name of [MapLayer]
+ * @param action action to execute per [MapObject] of the [MapLayer]
+ */
+fun TiledMap.forEachMapObject(layerName: String, action: (MapObject) -> Unit) {
+    layer(layerName).objects.forEach { action(it) }
+}
```
czyzby: Missing new line.

tiled/src/test/kotlin/ktx/tiled/mapLayer.kt
```diff
@@ -0,0 +1,33 @@
+package ktx.tiled
```
czyzby: Add `Test` suffix to the file name.

tiled/src/test/kotlin/ktx/tiled/mapLayer.kt
```diff
@@ -0,0 +1,33 @@
...
+        Assert.assertNotNull(customProperty)
+        Assert.assertEquals(123, customProperty)
+    }
+
+    @Test
+    fun `check if property from MapLayer exists`() {
+        Assert.assertTrue(mapLayer.containsProperty("active"))
+        Assert.assertFalse(mapLayer.containsProperty("x"))
+    }
+}
```
czyzby: Missing new line.

tiled/src/test/kotlin/ktx/tiled/mapLayer.kt
czyzby: Same goes for other tests.

tiled/src/test/kotlin/ktx/tiled/mapObjects.kt
```diff
@@ -0,0 +1,75 @@
...
+        assertEquals(polylineObject.polyline, polylineObject.shape)
+        assertEquals(polygonObject.polygon, polygonObject.shape)
+        assertEquals(Rectangle(0f, 0f, 1f, 1f), rectObject.shape)
+    }
+
+    @Test(expected = UnsupportedOperationException::class)
+    fun `retrieve shape from unsupported MapObject`() {
+        textureObject.shape
+    }
+}
```
czyzby: Missing new line.

tiled/src/test/kotlin/ktx/tiled/tiledMap.kt
```diff
@@ -0,0 +1,84 @@
...
+            it.isVisible = true
+        }
+
+        // also, test empty default layer which should do nothing
+        assertEquals(0, tiledMap.layer("non-existing").objects.count)
+        counter = 0
+        tiledMap.forEachMapObject("non-existing") { ++counter }
+        assertEquals(0, counter)
+    }
+}
```
czyzby: Missing new line.

Quillraven:
I have updated now all the things according to your feedback. 
The examples of the README.md should be also clearer now.

tiled/src/main/kotlin/ktx/tiled/mapLayer.kt
czyzby: For the sake of consistency, we can leave it as `containsProperty` and use `operator contains` for layers in tiled map.

czyzby:
@Quillraven I believe you have missed the comments that were automatically wrapped by GitHub. I've resolved some of the comments that were already addressed, please go through the entire review and check out the rest. Thanks!

czyzby:
To summarize the most important changes:
* Add a `MissingPropertyException` (and other similar exceptions, if necessary) that extends `GdxRuntimeException`.
* Add `property(name: String): T` extension methods that throw `MissingPropertyException` if the property is missing. Using a default value or handling nulls manually is not always desired.
* Do not return a default, static `MapLayer` instance if a layer with a particular name is missing. This behavior is error-prone.

Quillraven:
ah I see it now. Don't know why it didn't show me that before. Most likely I was blind  ;)

Will respond to it later today.

czyzby:
No problem. That's why I usually prefer the [changes view](https://github.com/libktx/ktx/pull/234/files), as it's harder to miss the comments.

tiled/src/main/kotlin/ktx/tiled/mapObject.kt
Quillraven: Unfortunately no. Those are three separate classes that contain a `MapProperties` attribute which is the container for properties.
The OOTB way to access those properties is by calling `.getProperties()` and then call `get`, `contains`, etc..

That's why I have to add them separately.

But maybe it would be a better approach to make them `MapProperties` extensions instead? Then it would be only once and instead of `property` or `containsProperty` you need to write `properties` or `properties.contains`.

What do you think?

czyzby: I would definitely add extensions for `MapProperties`. If you include the `operator` keyword in the `contains` extension method, it will also allow for the `"key" in map.properties` syntax.

As for the individual extensions to extract properties from `MapObject` and so on, I still think they are useful and we can leave them as they are. In other words, all of them could support the following methods:
- `property(key: String): T`: returns a property or throws an exception. Checks for nulls.
- `property(key: String, defaultValue: T): T`: returns a property or the default value. Checks for nulls.
- `propertyOrNull(key: String): T?`: returns a property or null if undefined.
- `containsProperty(key: String): Boolean`: true if property exists and is not null, false otherwise.

I'd leave `contains` without any suffix for checks specific to the map class. E.g. `TiledMap` could use operator `contains(layerName: String): Boolean` to verify if certain layers exist.

tiled/src/main/kotlin/ktx/tiled/mapObject.kt
Quillraven: From what I can tell from my maps at least and from the `com.badlogic.gdx.maps.tiled.BaseTmxMapLoader#loadObject` there are some default properties for every object.
Those are `x`, `y`, `width`, `height` and `id` (I did not know about width and height. Just noticed that now and will add them).

There would be other properties like `rotation` and `type` but they can be `null` and are therefore not added and I would not suggest an extension for them due to that.

That's why those properties should never be missing and the default value will never apply. 

I can add the exception extension if you want but personally - at least in my use cases - I prefer a default value instead of an exception.
E.g. I prefer to have an enemy created at 0,0 instead of not creating him at all. The first scenario gives me the hint that everything is working but something is wrong with the location where he is added. The second scenario could be anything since the enemy does not get created at all.
Also, in Kotlin there are no checked exceptions, correct? So a user would need to pro-actively add the try-catch block because the IDE will not warn him?

Again, I am openminded here. Just let me know what do you think is the better approach. Exception or default value - I will do as you wish ;)

tiled/src/main/kotlin/ktx/tiled/tiledMap.kt
Quillraven: You are right about the mutable part. It could be changed from the outside which is not nice.

The reason why I do it like this is because of my use-cases that I had. E.g. when I load a map I parse different layers and do something with them:
```kotlin
// this refers to a TiledMap
createSceneryEntities(this)
createEnemyEntities(this)
createSavepoints(this)
createNPCs(this)
createItemEntities(this)
createPortalEntities(this)
createTriggers(this)
```

One of these methods then looks like this:
```kotlin
private fun createSceneryEntities(map: Map) {
    map.mapObjects(LAYER_COLLISION).forEach { mapObj ->
        // loop through all map objects of that layer and
        // create the scenery entity according to the shape
        // of the map object.
        ecsEngine.scenery(world, mapObj.shape)
    }
}
```

So I don't care if the layer is there or not. If it is there then I will iterate over its objects and create collision scenery box2d objects out of it.
I would not like to have an exception because it is totally valid to leave out the collision layer.
Also, I would not like to have `isLayerExisting` calls everywhere. Granted, with Kotlin it is just a simple `?` character but especially when it comes to chaining I might end up with several `?`.

Of course this was my use-case and we should think here of a more generic approach for everyone. Just wanted to explain to you why I do it that way.

At least for me we can get rid of it since the `forEachMapObject` extension can take care of it.

I will make the change.

tiled/src/main/kotlin/ktx/tiled/tiledMap.kt
Quillraven: refer to `MapProperties` extension proposal from `MapObject`

Quillraven:
Added some comments to your remarks and changed a few things.

Main change would now be to make `MapProperties` extensions instead of having `MapObject`, `MapLayer` and `TiledMap` extensions. 

This would lead to changes like instead of `map.property("myProperty", "")` we would write `map.properties("myProperty", "")`. So basically property becomes properties ;)

What do you think?

tiled/src/main/kotlin/ktx/tiled/mapObject.kt
czyzby: Exceptions are easier to track and patch. With enemies randomly created at `0, 0`, you wouldn't be sure if the source of the bug is your spawning code or the map settings. With an exception thrown during map loading, you immediately know that the file is corrupted.

It's true that there are no checked exceptions, but then again - if the map files are correct, these kind of errors are not expected at all. You give the user the choice to handle it as they prefer.

czyzby: >  There would be other properties like rotation and type but they can be null and are therefore not added and I would not suggest an extension for them due to that.

You can add nullable properties for these.

czyzby:
I wouldn't expect a map of properties to be callable. If anything, I think square braces would be a standard syntax for this: `map.properties["myProperty"]`. If `get` currently requires a `Class` instance, we could certainly improve it with an operator `get` extension method.

czyzby:
Btw, @Quillraven, you didn't have to remove the `layer` utility altogether, I'd just prefer if it threw an exception in case of an unknown layer rather than return a default empty instance.

```kotlin
/**
 * Extension method to retrieve a non-null [MapLayer] of the [TiledMap]. In case the layer
 * cannot be found, a default empty layer is returned with name **ktx-default-map-layer**.
 * @param layerName name of [MapLayer]
 * @return [MapLayer] of given name or an empty default map layer, if the specified layer does not exist
* @throws UnknownLayerException if the given name is missing.
 */
fun TiledMap.layer(layerName: String) = layers[layerName] ?: throw UnknownLayerException(layerName)
```

tiled/src/main/kotlin/ktx/tiled/mapObject.kt
Quillraven: I tried to add an extension now for MapProperties like:

```kotlin
inline operator fun <reified T> MapProperties.get(key: String): T = get(key, T::class.java)
```

but it is shadowed and therefore ignored. So I think it is not possible to add `MapProperties` extensions?

So for now I will stick to TiledMap, MapObject and MapLayer extensions and will provide the functionality discussed above

Quillraven:
I have looked now into all open topics and adjusted the extensions accordingly. Also added the remaining missing standard properties, in case someone needs them.

I was not able to add extensions for `MapProperties` class due to shadowing issues which I don't know how to solve (maybe they are not solvable?).
Therefore, I kept the MapObject, MapLayer and TiledMap extensions.

Also added contains to TiledMap and the layer extension, this time throwing one of the newly added exceptions.

Let me know if this looks good then I will update the README.md accordingly. I did not want to do that now because I want to wait for the final version. Otherwise I will update it multiple times ;)

tiled/src/main/kotlin/ktx/tiled/mapObject.kt
czyzby: It would be shadowed only if there is a single parameter `get` method already.

Yeah, well, if there's no way to do it, stick to other extensions. Thanks for checking it out.

tiled/src/main/kotlin/ktx/tiled/exception.kt
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * [GdxRuntimeException] that is thrown when trying to access a non-existing [MapLayer] of a [TiledMap] instance
+ */
+class MissingLayerException(message: String) : GdxRuntimeException(message)
+
+/**
+ * [GdxRuntimeException] that is thrown when trying to access a shape of a [MapObject] that do not have any shape like [TextureMapObject]
+ */
+class MissingShapeException(message: String) : GdxRuntimeException(message)
```
czyzby: Missing new line.

tiled/src/main/kotlin/ktx/tiled/mapLayer.kt
```diff
@@ -0,0 +1,38 @@
...
+        ?: throw MissingPropertyException("Property $key does not exist for layer ${this.name}")
+
+/**
+ * Extension method to directly access the [MapProperties] of a [MapLayer]. The type is automatically
+ * derived from the type of the given default value. If the property is not defined the defaultValue will be returned.
+ * @param key property name
+ * @param defaultValue default value in case the property is missing
+ * @return value of the property or defaultValue if property is missing
+ */
+inline fun <reified T> MapLayer.property(key: String, defaultValue: T): T = this.properties[key, defaultValue, T::class.java]
```
czyzby: Can this call return a null? E.g. if the key is linked to an existing property with `null` value. It would make more sense to throw MissingPropertyException than receive a NPE. The same applies to other such calls.

tiled/src/main/kotlin/ktx/tiled/exception.kt
Quillraven: Out of curiosity, what is the problem if a new line is missing? Seems like auto-formatting does not add it automatically on my side.

tiled/src/main/kotlin/ktx/tiled/mapLayer.kt
Quillraven: In my opinion it cannot do that and it is also part of the Tests that I wrote. When checking the original method, from what I can tell, it will always return the defaultValue if either the key is missing or the linked value to the key is null.

```Java
public <T> T get (String key, T defaultValue, Class<T> clazz) {
	Object object = get(key);
	return object == null ? defaultValue : (T)object;
}
```

Quillraven:
Updated ReadMe now and also added the property extensions to Tile and TileSet. Those should be all objects that use the `MapProperties` class.

tiled/src/main/kotlin/ktx/tiled/exception.kt
czyzby: [Here's a few reasons](https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file). You can configure IntelliJ to [add it automatically](https://stackoverflow.com/questions/16761227/how-to-make-intellij-idea-insert-a-new-line-at-every-end-of-file).

tiled/src/main/kotlin/ktx/tiled/mapLayer.kt
czyzby: OK, thanks for verifitcation.

tiled/src/main/kotlin/ktx/tiled/mapObject.kt
Quillraven: changed the standard properties extensions as well to the version where it throws an exception instead of returning a default value

tiled/src/main/kotlin/ktx/tiled/tiledMap.kt
```diff
@@ -0,0 +1,152 @@
...
+        ?: throw MissingLayerException("Layer $layerName does not exist for map")
+
+/**
+ * Extension method to easily execute an action per [MapObject] of a given [MapLayer].
+ * If the layer does not exist then nothing is happening.
+ *
+ * @param layerName name of [MapLayer]
+ * @param action action to execute per [MapObject] of the [MapLayer]
+ */
+fun TiledMap.forEachMapObject(layerName: String, action: (MapObject) -> Unit) {
```
czyzby: This function can be inlined.

czyzby:
Looks OK now. I might do some minor documentation changes here and there, but I'm merging this module as is. Thank you!

czyzby:
@Quillraven Just a small tip on the tests. You don't have to store every object as a class property, especially if it's relevant only for a single test. It's OK to create objects inside the test functions.

I'm also a fan of the [given-when-then](https://en.wikipedia.org/wiki/Given-When-Then) testing structure. While Kotlin frameworks do not support this as keywords as e.g. Groovy's Spock does, it makes sense to compose your tests like that.

Basically, you divide your tests into 3 sections:
- given: preparation. You create and set up all necessary object here.
- when: execution. Ideally, a single tested method call.
- then: assertion. All of the checks that you need to perform.

You'll notice that a lot of the new and refactored tests in KTX are written like that.

Quillraven:
Ah I know about that because we do something similar for typescript tests at work.
Was not aware that this is actually a good practice. Will improve on that the nex time, thanks.

About the object creation... in this case I wanted to have it "global" because I wanted to reuse it amongst all tests but maybe that is not ideal. 

Thank you for your explanations and help. I learn from that  ;)

PR #237, opened 2019-12-22
removeAll/retainAll return Boolean; transfer less strict

cypherdare:
A couple of changes I hadn't pushed yet when #232 was merged.

`removeAll` and `retainAll` should return Boolean like the stlib functions do.

And I thought `transfer` could be less strict about the target array type.


czyzby:
Thanks, these will be available in the next release.

Issue #238, opened 2020-01-11
Missing touchDown / touchUp scene2d extensions?


Quillraven:
Hi,

I think that there are touchdown/touchup extensions missing in the actor ktx extensions or maybe that is on purpose.

What I am missing is a way to define a `touchDown` and `touchUp` callback for buttons where I want to react when they get pressed and released.

E.g. when you press the JUMP button the player should jump but when you release it he should cancel the jump and fall down again.

Click and Change did not work for me because click seems to only react when releasing the button and change only reacts on changes which means the first press is a trigger because the button gets "checked", releasing will not trigger anything and the next press will trigger it again.

What worked for me is the touchDown and touchUp events of the ClickListener.

The extension could look like this?

```kotlin
inline fun Actor.onTouch(
        crossinline touchDownListener: () -> Unit,
        crossinline touchUpListener: () -> Unit
    ): ClickListener {
        val clickListener = object : ClickListener() {
            override fun touchDown(event: InputEvent, x: Float, y: Float, pointer: Int, button: Int): Boolean {
                touchDownListener()
                return true
            }

            override fun touchUp(event: InputEvent?, x: Float, y: Float, pointer: Int, button: Int) = touchUpListener()
        }
        this.addListener(clickListener)
        return clickListener
    }
```

It could then be used like this

```kotlin
actor.onTouch(
                touchDownListener = { this.gameEventManager.dispatchInputKeyPressEvent(Key.JUMP) },
                touchUpListener = { this.gameEventManager.dispatchInputKeyReleaseEvent(Key.JUMP) }
            )
```

Would this be useful or do you suggest a different approach for such a use-case?

czyzby:
I believe this wasn't added, as `onClick` covers most use cases and this utility would require two lambdas or a boolean attribute with up/down state. We could add something similar to your suggestion.

czyzby:
Separate `onTouchDown` and `onTouchUp` extension methods could also do the job, although you'd add 2 separate listeners.

Quillraven:
I can look into it in more detail next weekend. I think having 2 separate listeners would not be a big issue as usually you do not have hundreds of those listeners so it will not have a big impact in my opinion.

Taken from the actors documentation:
```kotlin
import ktx.actors.*

label.onClick { 
  println("Label clicked!")
}

label.onClickEvent { inputEvent, actor ->
  // If you need access to the original InputEvent, use this expanded method variant.
  println("$actor clicked by $inputEvent!")
}
[...]

PR #239, opened 2020-01-19
(#238) adds touch event extensions for actors

Quillraven:
Adds touch event extensions for actors. I saw that the `InputEvent` actually has the information if a touchdown or touchup event is happening (InputEvent has a `type` which indicates that).

I did not use that at the moment but maybe it would be a better approach then the current one so that the user only specifies one lambda and in there he can add an if to either execute the down or up code.

Also, `touchDown` requires a boolean return type because only if `true` is returned then the `touchUp` for that actor will be called.
I don't know when you would return false but because of that the down lambda currently expects a boolean return type.

What do you think? Should we keep the extensions as they are right now (maximum flexibility) or should we simplify it?

### DIFF
--- a/actors/README.md
+++ b/actors/README.md
@@ -144,21 +144,27 @@ button.onTouchUp {
 
 button.onTouchEvent(
   // If you need access to the original InputEvent, use this expanded method variant.
-  downListener = { inputEvent, actor -> println("$actor down by $inputEvent!"); true },
+  downListener = { inputEvent, actor -> println("$actor down by $inputEvent!") },
   upListener = { inputEvent, actor -> println("$actor up by $inputEvent!") }
 )
+// or with a single lambda. In this case you can use InputEvent.Type to distinguish between touchDown and touchUp
+button.onTouchEvent( { inputEvent, actor -> println("$actor ${inputEvent.type} by $inputEvent!") })
 
 button.onTouchEvent(
   // If you need access to the local actor coordinates, use this expanded method variant.
-  downListener = { inputEvent, actor, x, y -> println("$actor down by $inputEvent! at ($x, $y)"); true },
-  upListener = { inputEvent, actor, x, y -> println("$actor up by $inputEvent! at ($x, $y)")}
+  downListener = { inputEvent, actor, x, y -> println("$actor down by $inputEvent at ($x, $y)!") },
+  upListener = { inputEvent, actor, x, y -> println("$actor up by $inputEvent at ($x, $y)!")}
 )
+// or again as single lambda
+button.onTouchEvent( { inputEvent, actor -> println("$actor ${inputEvent.type} by $inputEvent at ($x, $y)!") })
 
 button.onTouchEvent(
   // If you need access to the pointer and mouse button, use this expanded method variant.
-  downListener = { inputEvent, actor, x, y, pointer, mouseButton -> println("$actor down by $inputEvent! at ($x, $y) with pointer $pointer and mouseButton $mouseButton"); true },
-  upListener = { inputEvent, actor, x, y, pointer, mouseButton -> println("$actor up by $inputEvent! at ($x, $y) with pointer $pointer and mouseButton $mouseButton")}
+  downListener = { inputEvent, actor, x, y, pointer, mouseButton -> println("$actor down by $inputEvent at ($x, $y) with pointer $pointer and mouseButton $mouseButton!") },
+  upListener = { inputEvent, actor, x, y, pointer, mouseButton -> println("$actor up by $inputEvent at ($x, $y) with pointer $pointer and mouseButton $mouseButton!")}
 )
+// or again as single lambda
+button.onTouchEvent( { inputEvent, actor -> println("$actor ${inputEvent.type} by $inputEvent at ($x, $y) with pointer $pointer and mouseButton $mouseButton!") })
 ```
 
 Adding an `EventListener` which consumes typed characters:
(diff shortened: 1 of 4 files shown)


czyzby:
We could try overloads with a sensible defaults, so the users can omit the boolean.

actors/src/main/kotlin/ktx/actors/events.kt
```diff
@@ -80,6 +80,89 @@ inline fun <Widget : Actor> Widget.onClickEvent(
...
+  }
+  this.addListener(clickListener)
+  return clickListener
+}
+
+/**
+ * Attaches a [ClickListener] to this actor.
+ * @param downListener invoked each time this actor is touched. Consumes the triggered [InputEvent] and the [Actor] that
+ * the listener was originally attached to. Refer to [ClickListener.touchDown] for parameter details.
+ * * @param upListener invoked each time the touch of the actor is released. Consumes the triggered [InputEvent] and the [Actor] that
```
czyzby: Remove extra `* ` here and in some docs below.

actors/src/main/kotlin/ktx/actors/events.kt
```diff
@@ -80,6 +80,89 @@ inline fun <Widget : Actor> Widget.onClickEvent(
   return clickListener
 }
 
+/**
+ * Attaches a [ClickListener] to this actor.
+ * @param listener invoked each time this actor is touched.
+ * @return [ClickListener] instance.
+ */
+inline fun Actor.onTouchDown(crossinline listener: () -> Boolean): ClickListener {
```
czyzby: Could we overload this method to consume a lambda with `Unit` return type? This would simplify the API for the users.

actors/src/main/kotlin/ktx/actors/events.kt
```diff
@@ -80,6 +80,89 @@ inline fun <Widget : Actor> Widget.onClickEvent(
...
+ * the listener was originally attached to. Refer to [ClickListener.touchDown] for parameter details.
+ * * @param upListener invoked each time the touch of the actor is released. Consumes the triggered [InputEvent] and the [Actor] that
+ * the listener was originally attached to. Refer to [ClickListener.touchUp] for parameter details.
+ * @return [ClickListener] instance.
+ */
+inline fun <Widget : Actor> Widget.onTouchEvent(
+  crossinline downListener: (event: InputEvent, actor: Widget, x: Float, y: Float) -> Boolean,
+  crossinline upListener: (event: InputEvent, actor: Widget, x: Float, y: Float) -> Unit): ClickListener {
+  val clickListener = object : ClickListener() {
+    override fun touchDown(event: InputEvent, x: Float, y: Float, pointer: Int, button: Int): Boolean = downListener(event, this@onTouchEvent, x, y)
```
czyzby: You might want to move the part after return type definition to the next line to improve readability. Same goes for other longer methods.

Quillraven: Usually I use autoformat but you seem to have a different format then the "Kotlin standard" from intellij. Don't know how to take over your formatting.
But anyway, will move that to a new line :)

actors/src/main/kotlin/ktx/actors/events.kt
Quillraven: it would be possible yes and I would suggest to return true, otherwise touchup will never be called. 
I don't know if there is anyone out there who would return false for that method. TBH I don't really understand it. In my opinion in this context you always want to return true no matter what.

Shall I just do that to make it simpler?

actors/src/main/kotlin/ktx/actors/events.kt
czyzby: I think the XML with the configuration is in the repo, but it's not fully automatic - it doesn't break some lines that it should. I was planning on adding editor config and conforming to the official formatting guide, but that would break most of the Git history.

actors/src/main/kotlin/ktx/actors/events.kt
czyzby: > I don't know if there is anyone out there who would return false for that method.

I don't fully remember, but AFAIK it seems like this is either to stop other listeners from handling the event or to stop `touchUp` from firing. A niche use either way. If it's not trivial to add an overload, I think we should just go with the `Unit` lambda and let the users handle niche cases with custom listeners.

> With that we could also change `onTouchEvent` to only take one lambda and the user could the just check for `event.type` if it is touchDown or touchUp - what do you think?

Agreed, a simplified, single-lambda `onTouchEvent` could be useful.

czyzby: > I don't know if there is anyone out there who would return false for that method.

I don't fully remember, but AFAIK it seems like this is either to stop other listeners from handling the event or to stop `touchUp` from firing. A niche use either way. If it's not trivial to add an overload, I think we should just go with the `Unit` lambda and let the users handle niche cases with custom listeners.

> With that we could also change `onTouchEvent` to only take one lambda and the user could the just check for `event.type` if it is touchDown or touchUp - what do you think?

Agreed, a simplified, single-lambda `onTouchEvent` could be useful.

Quillraven: Added now also single lambda extensions for all three cases

actors/src/main/kotlin/ktx/actors/events.kt
Quillraven: Ah okay, I did notice that it does "something" but not 100% perfect. But I think it is fine and I hopefully was able to fix the formatting issues.

actors/README.md
```diff
@@ -131,6 +132,39 @@ label.onClickEvent { inputEvent, actor, x, y ->
   // If you need access to the local actor click coordinates, use this expanded method variant.
   println("$actor clicked by $inputEvent at ($x, $y)!")
 }
+
+button.onTouchDown {
+  println("Button down!")
+  true
```
czyzby: Out-of-date.

czyzby:
1 example in the README was not updated, other than that - it's ready to merge. I'll take care of this later today.

Issue #227, opened 2019-12-05
Ranges for randoms idea


cypherdare:
I make a lot of from-scratch particle effects for my live wallpapers (so I'm not limited by the GDX particle editor's features), and have found Kotlin ranges to be very helpful for very quickly tweaking values. I think this could be helpful anywhere you have something that spawns something in a game, whether it's a particle or enemy. Or for any procedural generator.

So I made some functions that make it even faster to tweak ranges. Operator functions that let you add or multiply constants by the endpoints of the range. Or defining a range by a center and tolerance. And then functions that return a random value from within the range.

For example, here's a simple version of something I've been working on.

```Kotlin
class ParticleLayer (val xStart: ClosedRange<Float>, 
        val yStart: ClosedRange<Float>, 
        val size: ClosedRange<Float> ) {
    private val particles = GdxArray<Particle>()
    //...

    fun update (delta: Float) { /* Updates particles and spawns them when necessary. */ }

    private fun spawnParticle() {
        val particle = particlePool.obtain()
            initialize(xStart.random(), yStart.random(), size.randomGaussian())
        particles.add(particle)
    }
}
```

I have multiple layers of particles, and the ones in the foreground and background are bigger to look out of focus (and move faster but I'm omitting that for the example).

So I first made one particle layer. As I tweak the values, instead of continually typing in new ranges, I can shift them around easily by adding a value:

```Kotlin
val layers = listOf(
    ParticleLayer(xStart = (-0.5f..0.5f) + 0.2f, // My particles needed to have their center shifted to look good
        yStart = 1.2f..2.5f,
        size = 1.3f..1.7f)
)
```

And after getting that looking good, I can copy-paste to get a new layer and then tweak the size range with simple multiplication:

```Kotlin
val layers = listOf(
    ParticleLayer(xStart = (-0.5f..0.5f) + 0.2f,
        yStart = 1.2f..2.5f,
        size = 1.3f..1.7f
    ),
    ParticleLayer(xStart = (-0.5f..0.5f) + 0.2f,
        yStart = 1.2f..2.5f,
        size = (1.3f..1.7f) * 1.3f
    )
)
```
 
Here are the functions I have in my personal library so far:

```Kotlin
fun ClosedRange<Int>.random() = MathUtils.random(endInclusive - start) + start
infix fun Int.amid(tolerance: Int) = (this - tolerance)..(this + tolerance)
fun ClosedRange<Float>.random() = MathUtils.random() * (endInclusive - start) + start
[...]

PR #240, opened 2020-01-23
Add ranges utitlities to math

cypherdare:
This is for #227.


math/src/main/kotlin/ktx/math/ranges.kt
```diff
@@ -0,0 +1,130 @@
...
+ * range.
+ *
+ * This function uses `MathUtils.randomTriangular(min, max, mode)`. Note that [endInclusive][ClosedRange.endInclusive]
+ * is treated as exclusive as it is not practical to keep it inclusive. Results are undefined for an empty range, and
+ * there is no error checking.
+ */
+fun ClosedRange<Float>.randomTriangular(normalizedMode: Float) =
+    MathUtils.randomTriangular(start, endInclusive,
+        normalizedMode * (endInclusive - start) + start
+    )
```
czyzby: Missing new line.

math/src/test/kotlin/ktx/math/rangesTest.kt
```diff
@@ -0,0 +1,244 @@
...
+  fun `should subtract from int range`() {
+    val original = -20..500
+    val subtrahend = 37
+    val shifted = original - subtrahend
+    assertEquals(original.first - subtrahend, shifted.first)
+    assertEquals(original.last - subtrahend, shifted.last)
+  }
+
+  @Test
+  fun `should produce expected random int values`() {
```
czyzby: Note sure about this test. It seems to go into too much detail about how the internal implementation works. For tests such as this, I think I'd repeat the `range.random` call, save the results and check:
- If every value is within the range.
- If the values (roughly) produce uniform distribution.

math/src/test/kotlin/ktx/math/rangesTest.kt
```diff
@@ -0,0 +1,244 @@
...
+  fun `should subtract from float range`() {
+    val original = -20f..500f
+    val subtrahend = 37f
+    val shifted = original - subtrahend
+    assertEquals(original.start - subtrahend, shifted.start)
+    assertEquals(original.endInclusive - subtrahend, shifted.endInclusive)
+  }
+
+  @Test
+  fun `should produce expected random float values`() {
```
czyzby: Same goes for this test. You can include an external math library in test scopes if that makes the asserts any easier.

math/src/test/kotlin/ktx/math/rangesTest.kt
```diff
@@ -0,0 +1,244 @@
...
+              val probability = 1f - leftTriangleArea - rightTriangleArea
+              checkRange to probability
+            }
+    for ((checkRange, probability) in rangesToExpectedProbabilities) {
+      var result = values.count { it in checkRange }.toFloat() / count
+      assert(abs(result - probability) / probability <= allowableError)
+    }
+  }
+
+}
```
czyzby: Missing new line.

cypherdare:
Good points, updated!

czyzby:
Thanks for the contribution.

PR #244, opened 2020-02-04
RenderScope() factory function

cypherdare:
Felt like this was missing. Not sure if it should be called `KtxScope()` for naming consistency, or if that makes it too vague.


czyzby:
Thanks, seems to be a useful feature. I'll update the CHANGELOG.

czyzby:
I went ahead and renamed this to `RenderingScope`. Thanks again for the contribution.

cypherdare:
You're welcome! Thanks for managing this great library.

PR #246, opened 2020-02-10
Make Context.register a non-inlined function…

serandel:
… so it can be mocked in tests and the bindings lambda stops being run


czyzby:
Thanks for the contribution, but it could break existing applications and cause (admittedly minor) performance decreases. I've outlined the reasons for closing this PR in #247.

PR #249, opened 2020-02-29
Range interpolation functions

cypherdare:
I find myself using these frequently when I use ranges for defining characteristics of my game objects. I also ran the formatter on rangesTest.kt.

### DIFF
--- a/math/src/test/kotlin/ktx/math/rangesTest.kt
+++ b/math/src/test/kotlin/ktx/math/rangesTest.kt
@@ -5,8 +5,6 @@ import com.badlogic.gdx.math.MathUtils
 import org.junit.Assert.*
 import org.junit.Test
 import kotlin.math.abs
-import kotlin.random.Random
-import kotlin.random.Random.Default
 
 /**
  * Tests [ClosedRange]-related utilities.
@@ -78,9 +76,9 @@ class RangesTest {
       .map { it.first until it.second }
       .forEach { innerRange ->
         val innerCount = values.count { it in innerRange }
-        assert(abs(innerCount.toFloat() - expectedCountEach) / expectedCountEach <= allowableError)
+        assertTrue(abs(innerCount.toFloat() - expectedCountEach) / expectedCountEach <= allowableError)
       }
-    assert(values.all { it in range })
+    assertTrue(values.all { it in range })
   }
 
   @Test
@@ -148,9 +146,9 @@ class RangesTest {
       .map { it.first..it.second }
       .forEach { innerRange ->
         val innerCount = values.count { it in innerRange }
-        assert(abs(innerCount.toFloat() - expectedCountEach) / expectedCountEach <= allowableError)
+        assertTrue(abs(innerCount.toFloat() - expectedCountEach) / expectedCountEach <= allowableError)
       }
-    assert(values.all { it in range })
+    assertTrue(values.all { it in range })
   }
 
   @Test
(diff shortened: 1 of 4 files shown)


math/src/test/kotlin/ktx/math/rangesTest.kt
```diff
@@ -232,4 +235,18 @@ class RangesTest {
     }
     assert(values.all { it in range })
   }
+
+  @Test
+  fun `should interpolate`() {
+    val start = Random.nextFloat()
+    val end = Random.nextFloat()
+    val progress = Random.nextFloat()
```
czyzby: Avoid using random values in tests, unless the seed is set beforehand.

In this case I'd use fixed values.

math/src/test/kotlin/ktx/math/rangesTest.kt
```diff
@@ -232,4 +235,18 @@ class RangesTest {
...
+    val range = start..end
+    val expectedLinear = (end - start) * progress + start
```
czyzby: Afaik it's exactly the same logic as the actual implementation, so it will likely return the same value. I generally avoid tests like this and rely on predefined values instead, e.g. start = 1, end = 3, progress = 0.5, expected result = 2.

math/src/test/kotlin/ktx/math/rangesTest.kt
```diff
@@ -232,4 +235,18 @@ class RangesTest {
...
+    assertEquals(range.lerp(progress), expectedLinear)
+
+    val interpolation = Interpolation.exp10
+    val expectedInterpolated = (end - start) * interpolation.apply(progress) + start
```
czyzby: Move this to a separate test, since it tests different functionality.

math/src/test/kotlin/ktx/math/rangesTest.kt
czyzby: The same applies to the test with interpolation.

cypherdare:
Good points. Thanks for reviewing.

math/src/test/kotlin/ktx/math/rangesTest.kt
```diff
@@ -232,4 +235,23 @@ class RangesTest {
...
+    assert(range.lerp(progress) in 262.5f amid 0.000001f)
+  }
+
+  @Test
+  fun `should interpolate`() {
+    val start = 0f
+    val end = 1f
+    val progress = 0.5f
+    val range = start..end
+    println(range.interpolate(progress, Interpolation.sineOut))
```
czyzby: Remove leftover print.

math/src/test/kotlin/ktx/math/rangesTest.kt
```diff
@@ -232,4 +235,23 @@ class RangesTest {
...
+    assert(range.interpolate(progress, Interpolation.sineOut) in 0.7071f amid 0.001f)
```
czyzby: You're using Kotlin's built-in `assert` instead of JUnit assertions. Beside the different error message, afaik this utility might have different behavior depending on the JVM flags:

> Throws an `AssertionError` if the value is `false` and runtime assertions have been enabled on the JVM using the `-ea` JVM option.

It might still work, but I'd prefer if you used `Assert.assertEquals`. I think it also provides a float overload with tolerance, so you won't need `amid`.

czyzby: Same applies to both tests.

cypherdare:
Oops! I went back and fixed the other ones from the previous ranges PR as well.

Issue #183, opened 2019-02-05
Integrate generic utilities from Cyberpunk framework


czyzby:
[Cyberpunk](https://github.com/ImXico/Cyberpunk) is an opinionated Kotlin LibGDX framework that aims to be on a little higher abstraction level than KTX. Some of its utilities are generic enough to make it into KTX. In particular, we should look into:

- [x] Screenshot utilities.
- [x] Profiling.
- [x] Camera utilities.
- [x] Image and text helpers.

See https://github.com/ImXico/Cyberpunk/issues/14. @ImXico

czyzby:
@ImXico I finally got around to move some utilities from Cyberpunk to KTX. I credited you in the contributors file, I hope that's OK. Thanks again for agreeing to do this!

czyzby:
Some feedback to consider:

> * Camera: from what I can tell it would be nice to also support a version without Vector2 and simply pass targetX and targetY
> * Text helpers: I wanted to use that for my floating texts but there is only a "String" version and I have a stringbuilder. Maybe there would be an option to also use stringbuilder? Second, and more important to me, it creates a Vector2 object. Since my texts are floating (=moving) I would create a vector2 object per frame. I would either need to redesign my code for this function or there should be an option to pass in a Vector2 object which will then contain the result.

Issue #182, opened 2019-02-05
Rewrite AssetStorage to the new ktx-async API


czyzby:
- [x] Add `ktx-assets-async` module.
- [x] `AssetStorage` rewrite.
- [x] Usage documentation.
- [x] Restore `ktx-freetype-async`, update to new API.
- [x] Restore `TextAssetLoader`, move to `ktx-assets`.

---

`AssetStorage` was removed during the `ktx-async` overhaul in `1.9.9-b1` due to reliance on old coroutines API and threading issues. The goal of this task is to rewrite `AssetStorage` in a fully thread-safe way, without reliance on specific coroutines dispatchers for asynchronous tasks.

amitkot:
I'm making changes to a project I created two years ago in which I used the previous version of `ktx-async` which depended on the experimental coroutines, and specifically `AssetStorage`.

I see that `AssetStorage` is no longer available in KTX.

Would you say the wise move right now is to switch to using `ktx-assets`'s `AssetManager` instead of `AssetStorage`?

czyzby:
@amitkot For now, yes. There were some issues with thread safety and unstable coroutines API usage, and I'm afraid `AssetStorage` needs a complete rewrite. Sorry.

czyzby:
@Quillraven @cypherdare @Jkly @maltaisn @jcornaz

I finally decided to rewrite the good old `AssetStorage`. While I still have to write usage examples and update `ktx-freetype-async`, I this point I'm happy with the `AssetStorage` implementation and its test suite.

To be honest, this is probably among top 5 most fun, challenging and complex pieces of code I've written. The initial implementation would definitely not pass the extensive tests I've written this time, including concurrent loading and unloading assets on multiple threads and coroutines.

I _think_ the current implementation is fully thread-safe and non-blocking. Unless the user of the API blocks the rendering or loading threads manually with `runBlocking` and whatnot, it should handle anything you throw at it without deadlocks or unexpected exceptions. I've run hundreds of thousands of concurrent loading and unloading requests of assets with nested dependencies - I encourage you to play with `should handle stress test` test in `storageTest.kt` to try to break it.

Anyway, I would appreciate it very much if you took the time to try `AssetStorage` out as a replacement for `AssetManager` or even go through the sources and do some code review. I know it's a huge pull request, but even superficial review would help a lot. If you're interested and have some time:
- [`ktx-assets-async`](https://github.com/libktx/ktx/tree/feature/assets-async/assets-async) - KTX module.
[...]

PR #255, opened 2020-03-22
Preferences API extensions

Quillraven:
Hi,

was working on preferences in my game and saw there is no nice kotlin extensions. I tried to come up with something but I was not sure how to solve `get` calls to avoid `getInt`, `getString`, etc..

Any ideas?

I will work on the documentation next week, if someone thinks this might be a useful addition. If not, feel free to delete this PR ;)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - **[FEATURE]** (`ktx-math`) Added `lerp` and `interpolate` extension functions for `Float` ranges.
 - **[FEATURE]** (`ktx-vis`) Added `image` (`VisImage`) factory methods consuming `Texture`, `TextureRegion` and `NinePatch`.
 - **[FEATURE]** (`ktx-preferences`) Added a new KTX module: Preferences API extensions.
+    - Added `set` operators for String, Int, Float, Long, Boolean, Pair<String, Any> and Any
+    - Added `get` operator
+    - Added `flush` extension function that takes additional `Preferences` calls as a Lambda
 
 #### 1.9.10-b4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - **[FEATURE]** (`ktx-async`) Added `RenderingScope` factory function for custom scopes using rendering thread dispatcher.
 - **[FEATURE]** (`ktx-math`) Added `lerp` and `interpolate` extension functions for `Float` ranges.
+- **[FEATURE]** (`ktx-preferences`) Added a new KTX module: Preferences API extensions.
 
 #### 1.9.10-b4
 
--- a/preferences/README.md
+++ b/preferences/README.md
@@ -1,0 +1,9 @@
+[![Maven Central](https://img.shields.io/maven-central/v/io.github.libktx/ktx-collections.svg)](https://search.maven.org/artifact/io.github.libktx/ktx-collections)
+
+# KTX: preference utilities
+
+Utilities and extension function for LibGDX preferences.
+
+### Why?
+
+TBD
--- a/preferences/README.md
+++ b/preferences/README.md
@@ -7,3 +7,57 @@ Utilities and extension function for LibGDX preferences.
 ### Why?
 
 TBD
+
+```kotlin
+import com.badlogic.gdx.Gdx
+import ktx.preferences.flush
+import ktx.preferences.get
+import ktx.preferences.set
+
+private class Player(val name: String = "Player 1", val life: Int = 100)
+
+fun main() {
+  val prefs = Gdx.app.getPreferences("MyPreferences")
+
+  // set values with new set operator functions
+  prefs["Key1"] = "someStringValue"
+  prefs["Key2"] = 1
+  prefs["Key3"] = 1f
+  prefs["Key4"] = 1L
+  prefs["Key5"] = true
+  // Any classes are automatically converted to a Json string
+  prefs["Key6"] = Player()
+
+  prefs.run {
+    // use Pairs to update preference values
+    set("Key7" to 123)
+    set("Key8" to "someOtherStringValue")
+  }
+
+  // get values with new get operator function
+  val value1: String = prefs["Key1"]
+  val value2: Int = prefs["Key2"]
+  val value3: Float = prefs["Key3"]
+  val value4: Long = prefs["Key4"]
+  val value5: Boolean = prefs["Key5"]
+  // Any classes are automatically loaded from a Json string
+  val value6: Player = prefs["Key6"]
+  val value7: Int = prefs["Key7"]
+  val value8: String = prefs["Key8"]
+
+  println(value1)
+  println(value2)
+  println(value3)
+  println(value4)
+  println(value5)
+  println("${value6.name} - ${value6.life}")
+  println(value7)
+  println(value8)
+
+  // adjust preferences before calling flush
+  prefs.flush {
+    set("Key9" to 10000)
+    set("Key10" to true)
+  }
+}
+```
--- a/preferences/README.md
+++ b/preferences/README.md
@@ -6,7 +6,21 @@ Utilities and extension function for LibGDX preferences.
 
 ### Why?
 
-TBD
+LibGDX [Preferences](https://github.com/libgdx/libgdx/wiki/Preferences) do not support a generic way to
+set and get values. Since they work very similar to a Map, they should ideally support a similar
+syntax. Especially with Kotlin we can use the advantage of square bracket operators.
+
+### Guide
+
+- Values can be set via new `set` operators using the square bracket syntax. It is no longer needed
+to call type specific methods like `putString` or `putBoolean`. `Set` already supports any type.
+- Values can be retrieved via new generic `get` operator.
+- Preferences now support `Pair<String, Any>` parameter. Values can be set by using the `infix to` function
+- New `flush` extension that supports a Lambda to easily update the Preferences before flushing.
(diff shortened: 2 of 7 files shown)


preferences/src/main/kotlin/ktx/preferences/preferences.kt
```diff
@@ -0,0 +1,18 @@
...
+
+operator fun Preferences.set(key: String, value: Boolean): Preferences = this.putBoolean(key, value)
+
+operator fun Preferences.set(key: String, value: Int): Preferences = this.putInteger(key, value)
+
+operator fun Preferences.set(key: String, value: Long): Preferences = this.putLong(key, value)
+
+operator fun Preferences.set(key: String, value: Float): Preferences = this.putFloat(key, value)
+
+fun Preferences.flush(operations: Preferences.() -> Unit) {
```
czyzby: This is a good candidate for an `inline` function.

preferences/README.md
```diff
@@ -0,0 +1,9 @@
+[![Maven Central](https://img.shields.io/maven-central/v/io.github.libktx/ktx-collections.svg)](https://search.maven.org/artifact/io.github.libktx/ktx-collections)
+
+# KTX: preference utilities
+
+Utilities and extension function for LibGDX preferences.
+
+### Why?
+
+TBD
```
czyzby: TODO.

preferences/src/test/kotlin/ktx/preferences/preferencesTest.kt
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.Test
+import org.mockito.Mockito
+
+class PreferencesTest {
+  private lateinit var preferences: Preferences
+  private val map = ObjectMap<String, Any>()
+
+  @Before
+  fun `setup preferences`() {
+    preferences = Mockito.mock(Preferences::class.java)
```
czyzby: Since you mock (nearly?) the entire class, you might as well implement `Preferences` in tests. Should be easier to read than a bunch of mocks.

czyzby:
I've thought about this and I think this could be added as a separate module. If there are tiny modules with pretty general purpose utilities like `ktx-log`, I don't see why preference utilities couldn't be separated as well.

Quillraven:
Made a quick update on the PR with reified set and get. Also thought of a way to maybe use the pair syntax but did not come up with a very nice way :D 

I also replaced the mock now with an implementation as you suggested.

Will update Readme by end of this week.

czyzby:
Overloads for `set` were fine, you can bring them back. Just leave one generic function for JSON serialization of properties. It's `get` that's problematic and needs the reified types.

preferences/src/main/kotlin/ktx/preferences/preferences.kt
```diff
@@ -0,0 +1,43 @@
+package ktx.preferences
+
+import com.badlogic.gdx.Preferences
+import com.badlogic.gdx.utils.GdxRuntimeException
+
+inline operator fun <reified T> Preferences.set(key: String, value: T): Preferences {
```
czyzby: Please revert the previous overloaded setters, this is basically a workaround that we should use for `get` only due to no union types.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - **[FEATURE]** (`ktx-async`) Added `RenderingScope` factory function for custom scopes using rendering thread dispatcher.
 - **[FEATURE]** (`ktx-math`) Added `lerp` and `interpolate` extension functions for `Float` ranges.
+- **[FEATURE]** (`ktx-preferences`) Added a new KTX module: Preferences API extensions.
```
czyzby: Typically new modules list all of their features. You can omit that for now if you want, I can add that once the PR is accepted.

preferences/src/main/kotlin/ktx/preferences/preferences.kt
```diff
@@ -0,0 +1,43 @@
...
+}
+
+inline operator fun <reified T> Preferences.get(key: String): T {
+  return when (T::class) {
+    String::class -> this.getString(key) as T
+    Boolean::class -> this.getBoolean(key) as T
+    Int::class -> this.getInteger(key) as T
+    Float::class -> this.getFloat(key) as T
+    Long::class -> this.getLong(key) as T
+    else -> throw GdxRuntimeException("Unsupported ${T::class}")
```
czyzby: Do we want to throw an exception or automatically convert property to JSON? I could see it being useful for more complex property objects.

Quillraven: should we then always create a new json instance ? or make it a parameter with default value and in case someone already has an instance, he can pass it?

Quillraven: should we then always create a new json instance ? or make it a parameter with default value and in case someone already has an instance, he can pass it?

Quillraven: should we then always create a new json instance ? or make it a parameter with default value and in case someone already has an instance, he can pass it?

czyzby: Saving preferences is not something you do every frame. Adding more parameters only brings complexity and unclear syntax - after all, you don't need the `Json` instance if all you do is save basic properties. I think we should create a new instance and document this in the comment.

Quillraven:
I have merged the develop branch now and resolved the CHANGELOG.md conflicts. Also, finished the documentation and everything else from my POV.

Let me know if there is anything to change :)

czyzby:
Thanks, I believe that's all as well. I'll do some minor refactoring and publish the module snapshot today. Thank you for your contribution!

czyzby:
@Quillraven I ended up things a bit as I looked closely into the code and the tests:

* `get` now returns a nullable value, since the preference might be missing. A variant with default value was added. The tests were expanded with examples of missing preferences.
* `Double` values did not work as expected with `get`/`set` (they were treated as JSON-serializable objects). Since this is a pretty commonly used basic data type, I've added some special measures to inform the users that it's not supported by `Preferences`. I did not do the same for `Short` and `Byte` though.
* I've split the code examples in the `README` into small sections.

You're welcome to [review the changes](https://github.com/libktx/ktx/commit/addd74642c299243071e0a79b629fbe83e90c18a).

czyzby:
@Quillraven I think there were quite a few useful utilities added since the last release, and this module is a major contribution. We might publish a new stable version this week. Let me know if you think something is missing or if you want to set up any more pull requests soon.

Quillraven:
looks good to me and I like the way you tested the flush extension. I always learn something from your changes, thanks!

Found a small typo but other than that it looks great.

I have no plans for other PRs as I am very busy at work for the next couple of weeks haha.

Looking forward to the new release - thank you a lot!

czyzby:
> Found a small typo but other than that it looks great.

Which file and line?

Quillraven:
51 in preferences.kt

I thought I added a comment but now I don't see it anymore. You wrote `amd`
instead of `and`.

MJ <notifications@github.com> schrieb am Mi., 25. März 2020, 20:34:

czyzby:
@Quillraven Thanks, good catch. (Also, Intel probably hates me.)

czyzby:
@Quillraven By the way, snapshot release should already be out. Assuming you're using the preferences extensions in a personal project, I'd appreciate it if you switched to KTX and tested the API - especially around the JSON serialization.

Quillraven:
will work on the private project over the weekend and will update to ktx preferences.

If something comes up, I will let you know  :)

czyzby:
@Quillraven Just wanted to let you know that I added some extra utilities from the Cyberpunk framework (#183), including taking screenshots, moving camera and profiling. Snapshots with the changes should already be out. If you have some free time for testing, I'd appreciate it if you looked through the code and tried it out in your project. :)

Quillraven:
@czyzby I updated now to the snapshot, kotlin 1.3.71 and gradle 6.3 (to use the exclusiveContent functionality for the snapshot ;) )

Looked into the Cyberpunk extensions but I cannot use any of them, sorry:
1) Screenshot: sounds useful but I don't use screenshots and no plans to support that for the time being
2) Profiling: same as above and my code is anyway perfectly optimized :P
3) Camera: from what I can tell it would be nice to also support a version without Vector2 and simply pass targetX and targetY
4) Text helpers: I wanted to use that for my floating texts but there is only a "String" version and I have a stringbuilder. Maybe there would be an option to also use stringbuilder? Second, and more important to me, it creates a Vector2 object. Since my texts are floating (=moving) I would create a vector2 object per frame. I would either need to redesign my code for this function or there should be an option to pass in a Vector2 object which will then contain the result.

This is just my personal opinion from my current project. I think in general they are useful extensions but at the moment I cannot use them, sorry.

czyzby:
@Quillraven Thanks for the feedback, it's actually very helpful.

By the way, I'm currently working on #182 (about time). I still have a bunch of tests to write and refactor, but the basic functionality is there. I should be done by tomorrow, unless there's some major bug that I'll have to fix. I've written some basic mock-up apps loading and using some assets, but if you're currently using `AssetManager`, I'd appreciate if you tested this one as well. I'll let you know when I'm done.

Quillraven:
sure just let me know. I am using assetmanager in the loading screen that actually loads all assets currently since it is still a small game and does not need load/unload mechanics.

czyzby:
@Quillraven Just wanted to let you know that `1.9.10-b5` was released with your preferences module and the new `AssetStorage` implementation.

Quillraven:
thank you! will update my project over the weekend :)

Issue #270, opened 2020-04-16
Additional ktx-ashley extensions


Quillraven:
From what I know it is a best practice in ashley to remove entities at the end of a frame or at the beginning, when using a `PooledEngine` because otherwise you sometimes get strange behaviors when removing and adding entities within the same frame.

Therefore, I use a `RemoveComponent` in most of my projects that is added to an entity to mark it as being removed. Some `IteratingSystems` then exclude entities with a `RemoveComponent`. A `RemoveSystem` is then added at the end or at the front of all engine systems which then removes such entities from the engine.

There is no kotlin extension for that, from what I have seen and that's why you have to write something like 

```kotlin
entity.add(engine.createComponent(RemoveComponent::class.java))
```

Might be a good idea to have the existing `with` function from the creation DSL for such things?

```kotlin
entity.with<RemoveComponent>
```

or something similar?

Also, very rarely you need to access certain systems from the engine. Here, I also could not find a kotlin counterpart and the current code is like:

```kotlin
engine.getSystem(SomeSystem::class.java)
```

Would it be possible to have something like:

```kotlin
engine[RemoveSystem::class]
// or
engine<RemoveSystem>
```

I could try something out on the weekend and provide a PR. What would be the preferred solutions?

czyzby:
I'll have to review Ashley module to give you more insights. You're welcome to submit the PR, if you already have a prototype of these features.

@Jkly Thoughts?

czyzby:
@Quillraven 

`Entity` has no access to `engine` after creation, so we would either have to return `EngineEntity` (which has the `with` extension for adding components) or add a utility method such as `Entity.addComponent<Type>(engine)`.

```kotlin
inline fun <reified C: Component> Entity.addComponent(engine: Engine, configure: (@AshleyDsl C).() -> Unit): C {
  val component = engine.create(configure)
  add(component)
  return component
}

// Usage example:
class Transform(var x: Float = 0f, var y: Float = 0f) : Component
fun example(entity: Entity, engine: Engine) {
  entity.addComponent<Transform>(engine) {
    x = 10f
    y = 10f
  }
}
```

As for the system, we could add something like this:

```kotlin
inline fun <reified S: EntitySystem> Engine.getSystem() = getSystem(S::class.java)

// Usage:
engine.getSystem<MySystem>()
```

If you'd like to refine my examples and write the tests, you're welcome to submit a PR. Otherwise I'll look into it during the weekend.

Quillraven:
[...]

Issue #269, opened 2020-04-16
Rewrite VisUI DSL module


czyzby:
- [x] Extend `ktx-scene2d` module in `ktx-vis`.
- [x] Unify `ktx-vis` and `ktx-scene2d` APIs and tests.
- [x] Replace `VisDsl` marker with `Scene2dDsl` marker.
- [x] Inline `ktx-vis` factory methods.
- [x] Improve `ktx-vis` documentation.

---

`ktx-scene2d` was originally added by me, while around the same time `ktx-vis` was created by the VisUI author. While the original plan was to either make `ktx-vis` extend `ktx-scene2d` or to have both extend a common GUI building module (#9), they remained separate due to different approaches and lack of crucial features at the time (e.g. default values for lambda parameters in inlined methods). As an example, `ktx-scene2d` originally had 2 or 3 factory methods for each actor - with inlined building lambda, without the lambda parameter and optionally top-level factory for parental actors. Such code duplication was unacceptable in `ktx-vis`, since VisUI had a lot more actors to cover.

Since then, both Kotlin features and `ktx-scene2d` API have improved and stabilized (#46, #48, #94, #268). Especially after #268, `ktx-vis` approach will feel vastly different to `ktx-scene2d`, which makes the transition between the two more difficult. Besides, `ktx-vis` lacks some useful features provided by `ktx-scene2d`:
- inlined factory methods for improved performance;
- simple creation of top-level actors that are not parental;
- complete documentation of all factory methods and their parameters.

The goal of this issue is to unify both modules, so that `ktx-vis` would extend `ktx-scene2d` and provide a similar set of features, rather than remain an alternative.

czyzby:
@kotcrab Just a heads up: I finally unified the GUI modules. I believe we didn't do this in the first place because there were no default values for lambda parameters in inlined methods, among other things. Anyway, if you want to take a look and check if anything is missing (maybe something from the latest VisUI versions), you can find the latest code on the [`develop`](https://github.com/libktx/ktx/tree/develop/vis) branch

kotcrab:
Nice! Honestly I don't remember which VisUI version was this originally based on but I doubt anything important is missing. Most [VisUI changes](https://github.com/kotcrab/vis-ui/blob/master/ui/CHANGES.md) were bug fixes after certain point.

Issue #268, opened 2020-04-16
Scene2D DSL refactoring to ease creation of actors without parents


czyzby:
- [x] Add `scene2d` object with access to Scene2D DSL.
- [x] Remove `KWidget.appendActor`.
- [x] Rename `addTooltip` and `addTextTooltip` to `tooltip` and `textTooltip` respectively.

---

`ktx-scene2d` currently defines top-level factory functions for commonly used parental actors, e.g. `table` or `window`. Creating these widgets with KTX DSL is straightforward:

```kotlin
val myTable = table {
  // Define children here.
  label("Text")
}
```

A common use case that is currently unsupported is creating non-parental actors with the DSL:

```kotlin
val myLabel = label("Text") {
  // Compilation error: top-level factory function for label is not available.
}
```

We could add similar factory functions for all LibGDX widgets, but it leads to 2 issues:
- Code duplication: we now have to maintain and test 2 factory functions for each widget.
- Name collisions: since both nested `label` children and top-level `label` widgets are now created with methods with the same name, it's easy to include the wrong one (e.g. via code completion), which leads to unexpected behaviour (e.g. label not added to the parent).

Instead, a unified API for defining root actors - regardless of whether they can or cannot have children - will be provided, using the same set of factory methods. To achieve that, `scene2d` object will be created, implementing the `KWidget` interface to have complete access to the Scene2D DSL. Additionally, it would provide factory methods for widgets that are normally never used as children, such as `window` and `dialog`. For example:

```kotlin
val myTable = scene2d.table {
  // Children definitions remain the same:
  label("Text")
}
val myLabel = scene2d.label("Text") {
  // Now it works!
}
```

Migration to the new API would be very simple, as all that needs to be done is appending the `scene2d.` prefix to the root actor definitions. In the long run, the usability of the module will improve, as now it would be simpler to define parent-less actors (such as labels) and name collisions are no longer an issue.

Existing top-level factory functions would be deprecated and eventually removed in the following release.

Additionally, the following changes will be applied:
- `addTextTooltip` and `addTooltip` use non-standard method names and will be renamed to `tooltip` and `textTooltip` to match other widgets.
- Internal utility method `KWidget.appendActor` is no longer necessary since Kotlin 1.2 introduced default values for lambda parameters in inlined functions. Since this is a part of the internal API, it will be removed without prior deprecation.

czyzby:
[...]

Issue #43, opened 2017-04-15
Migrate unit tests to a Kotlin testing framework


czyzby:
[KotlinTest](https://github.com/kotlintest/kotlintest) is a promising testing framework that we can use to replace the current non-idiomatic JUnit tests. [Mockito-Kotlin](https://github.com/nhaarman/mockito-kotlin) greatly improves mocking API and should be used instead of vanilla Mockito.

czyzby:
Tests written in KotlinTest cannot be run individually. You can run individual test _suite_, not but an individual _test_.  It's not that tedious, since most KTX tests take milliseconds to run, but it's certainly a step backwards from JUnit. I think using JUnit with Kotlin backticks to name test methods, `Mockito-Kotlin` for mocks and KotlinTest assertions is optimal for now.

Jkly:
Did you consider using Spek? It's more idiomatic to Kotlin than JUnit, and with the https://github.com/raniejade/spek-idea-plugin you are able to run each test individually.

czyzby:
I've seen this framework a while back and dismissed it for some reason. I think it required a lot of extra setup and JUnit 5 usage. I'll look into it again, thanks.

Anyway, Box2D utilities is something I was planning on eventually adding to KTX - and I really don't like reinventing the wheel. Would you consider merging your `gdx-box2d-kotlin` repository with KTX as an additional `ktx-box2d` module? I'd have to review the code first and might end up refactoring some stuff, but the README example is promising and looks pretty much like idiomatic Kotlin builder API.

Jkly:
Sure, I'd be more than happy to contribute `gdx-box2d-kotlin` - I could prepare a PR and let you review and refactor the code that way? (or if you just want to dive in and grab what you want, I'm ok with that too)

czyzby:
Thanks, that's great. It's possible to merge 2 repositories keeping the commits history, I can do that on a separate branch. You'll appear as contributor and all of your commits will be there.

Can't promise anything, but I'll try to look into it this week. I'll create a separate issue for that.

czyzby:
#52

czyzby:
After a quick look, I think you're overusing field delegation and Box2D class wrappers.

For example, you can add `body` method as an extension of `World`:
```Kotlin
fun World.body(init: (@Box2dDsl FixtureDef).() -> Unit) {
}
```
This wouldn't force the users to wrap and unwrap Box2D classes with your DSL objects.

You overuse delegation mechanism:
[...]

PR #272, opened 2020-04-17
ktx-ashley: Engine.getSystem, Engine.get and Entity.add

Quillraven:
As discussed in #270 I provided now three additions for ktx-ashley

Out of curiosity: it seems like a different testing framework was used for the ashley module? What was the reason to not use JUnit?

edit: seems like build failed but not related to the ashley module. Did I do something wrong with merging develop branch into my branch? :(


CHANGELOG.md
```diff
@@ -54,6 +54,9 @@ Note that the actors can still be created via standard DSL. See the migration gu
   - `styleName` parameters in factory methods were renamed to `style` for consistency with `ktx-scene2d`.
   - `@VisDsl` DSL marker is replaced with `@Scene2dDsl` marker from `ktx-scene2d`.
   - The sources documentation was greatly expanded.
+ - **[FEATURE]** (`ktx-ashley`) 
+   - Added `Engine.get` operator and `Engine.getSystem` function to access an `EntitySystem`
+   - Added `Entity.add` function to create a `Component` for an already existing `Entity`
```
czyzby, line 59: Add these as separate `- **[FEATURE]**` lines. I usually only merge multiple things into a single bullet point if it is a major feature consisting of multiple changes. These are completely separate extensions.

ashley/README.md
```diff
@@ -17,11 +17,13 @@ builder DSL.
 `ktx-ashley` provides the following extensions and utilities:
 
 - `Engine.add` and `Engine.entity` extension methods provide type-safe building DSL for creating `Entities`.
+- `Engine.getSystem` and `Engine.get` (`[]` operator) to access an `EntitySystem` of the engine
```
czyzby, line 20: Missing dot, same goes for the other line.

ashley/src/main/kotlin/ktx/ashley/engines.kt
```diff
@@ -74,6 +75,22 @@ inline fun Engine.entity(configure: EngineEntity.() -> Unit = {}): Entity {
   return entity
 }
 
+/**
+ * Quick [EntitySystem] retrieval.
+ *
+ * @return the [EntitySystem] of the given type. May be null if it does not exist.
+ * @see Engine.getSystem
+ */
+inline fun <reified T : EntitySystem> Engine.getSystem(): T? = getSystem(T::class.java)
```
czyzby, line 84: I could see it being more useful if it returned an `EntitySystem` or threw an exception. You usually know which systems you add to an engine.

`get` could remain nullable, so we support both options and these extensions are not redundant.

ashley/src/main/kotlin/ktx/ashley/engines.kt
```diff
@@ -74,6 +75,22 @@ inline fun Engine.entity(configure: EngineEntity.() -> Unit = {}): Entity {
...
+
+/**
+ * Quick [EntitySystem] retrieval.
+ *
+ * @return the [EntitySystem] of the given type. May be null if it does not exist.
+ * @see Engine.getSystem
+ */
+inline operator fun <reified T : EntitySystem> Engine.get(type: KClass<T>): T? = getSystem(type.java)
```
czyzby, line 92: I don't think you need to inline this or use a reified type if you pass a `type` parameter. It will be a little bit more useful, since you can't dynamically call inlined methods with arbitrary types.

ashley/src/main/kotlin/ktx/ashley/entities.kt
```diff
@@ -65,3 +66,15 @@ operator fun <T : Component> Entity.contains(mapper: ComponentMapper<T>): Boolea
...
  */
 inline fun <reified T : Component> Entity.remove(): Component? = remove(T::class.java) as? T
+
+/**
+ * Adds a [Component] to this [Entity]. If a [Component] of the same type already exists, it'll be replaced.
+ *
+ * @return the Entity for easy chaining
+ * @see Entity.add
+ */
+inline fun <reified T : Component> Entity.add(engine: Engine, configure: (@AshleyDsl T).() -> Unit = {}): Entity {
```
czyzby, line 76: Why is `Entity` returned by this method? I understand that Ashley does this for chaining, but in their case you already have access to the `Component` instance. In this case we're creating a new `Component`, so it makes sense to return it. I think the return type should be `T`.

Consider renaming this to `addComponent`. If the reified type is extracted e.g. from variable, the method name becomes confusing as it reads as "add engine". For example:

```kotlin
val transform: Transform = entity.add(engine)
```

ashley/src/test/kotlin/ktx/ashley/EnginesSpec.kt
```diff
@@ -124,5 +124,16 @@ object EnginesSpec : Spek({
         assertThat(entity.getComponent(Texture::class.java)).isNotNull()
       }
     }
+
+    describe("getSystem function") {
+      it("should add a system and return it") {
+        engine.addSystem(TestSystem())
+        assertThat(engine.getSystem<TestSystem>()).isNotNull()
```
czyzby, line 131: Instead of checking if the system is null, you can check if it matches the exact instance of `TestSystem`.

You have only tested the "happy path". You should also add 2 tests for `getSystem` and `get` when the system is absent in the engine.

ashley/src/test/kotlin/ktx/ashley/EntitiesSpec.kt
```diff
@@ -81,5 +85,12 @@ object EntitiesSpec : Spek({
         assertThat(Texture.mapper !in entity).isTrue()
       }
     }
+
+    describe("add component function") {
+      it("should add a component and return the entity") {
+        assertThat(entity.add<Transform>(engine)).isEqualTo(entity)
```
czyzby, line 91: Check if component is returned, see other comments.

ashley/src/test/kotlin/ktx/ashley/EntitiesSpec.kt
```diff
@@ -81,5 +85,12 @@ object EntitiesSpec : Spek({
         assertThat(Texture.mapper !in entity).isTrue()
       }
     }
+
+    describe("add component function") {
```
czyzby, line 89: Add another test with a configuration block for `Transform` to check if the changes are correctly applied (i.e. change `x` and `y`  with `add(engine) { ... }` and check their values).

czyzby:
Don't worry about the 1.11 build, it seems that it is an indeterministic test (concurrency magic). 4 threads were expected to be utilized, but 3 were sufficient to execute the coroutines. I'll fix the check to be less strict.

ashley/src/main/kotlin/ktx/ashley/engines.kt
Quillraven: Totally agree. When you get a system you usually know that it is there and you want to do something with it (except you messed up something ;) ).

I will throw an exception.

ashley/src/main/kotlin/ktx/ashley/engines.kt
Quillraven: I tried it without reified but then you lose the system type information and you just get a generic EntitySystem. But maybe I am missunderstanding something?

czyzby:
Try merging with the `develop`, the tests should pass as expected.

ashley/src/main/kotlin/ktx/ashley/engines.kt
czyzby: This compiles for me, you might have confused it with the `getSystem` extension:

```kotlin
operator fun <T : EntitySystem> Engine.get(type: KClass<T>): T? = getSystem(type.java)
```
image

ashley/src/main/kotlin/ktx/ashley/entities.kt
Quillraven: I thought it makes sense to have the same behavior as the original ashley function and returning the entity does not really help a lot because you anyway can specify component stuff in the lambda?

currently you could do something like:
```kotlin
entity.add<Damaged> {
  damage = 3f
}.add<Knockback> {
  force = 2.5f
}
```

When you return the component, you can no longer chain except you use kotlin's apply/run/with/let.

What do you think? Shall I still change it?

czyzby:
> Out of curiosity: it seems like a different testing framework was used for the ashley module? What was the reason to not use JUnit?

The module was written around the time where we were considering using other Kotlin-centric testing libraries (#43), and Jkly has written the tests with Spek. I find these tests pretty easy to read, as I'm used to similar approach in other languages (this looks very familiar to JS/TS tests), but they might have a steeper learning curve for Java developers than JUnit, so I decided against using it in other modules.

ashley/src/main/kotlin/ktx/ashley/entities.kt
czyzby: I don't find chaining particularly cleaner than:

```kotlin
entity.apply {
  add<Damaged> {
    damage = 3f
  }
  add<Knockback> {
    force = 2.5f
  }
}
```

I think chaining methods are usually a legacy from Java that make up for the lack of DSL capabilities.

Consider that if you need a reference to the created component, you'd have to do this:
```kotlin
lateinit var damaged: Damaged
entity.add<Damaged> {
  damage = 3f
  damaged = this
}
```

Or this:
```kotlin
entity.add<Damaged> {
  damage = 3f
}
val damaged = entity.get<Damaged>()
```

Instead of this:
```kotlin
val damaged = entity.add<Damaged> {
  damage = 3f
}
```

As a rule of thumb - if a utility method creates something, we want a reference to it. We already have a reference to the entity, but we might need a reference to the component we have just created.

Quillraven:
I have never seen tests in that format but I personally also prefer them over JUnit from what I can tell so far. Will try them out in a new project, thanks for the info!

I have applied the changes now. Please look over it again and let me know if it is okay now.

Quillraven:
is this already part of a snapshot release?

czyzby:
Yes, now snapshot releases are automatic. Everything merged to develop is deployed as snapshot to Maven Central. If anything is missing, let me know - that would be a bug.

Quillraven:
@czyzby I don't have that in my SNAPSHOT. Can you check on your site please? Maybe caching problem on my end?

czyzby:
@Quillraven Snapshots are not hosted on my site, it's the [official snapshot repository](https://oss.sonatype.org/content/repositories/snapshots/io/github/libktx/). Using the `--refresh-dependencies` flag usually works, although [there are some alternatives](https://stackoverflow.com/questions/32652738/how-can-i-force-update-all-the-snapshot-gradle-dependencies-in-intellij/47557700).

By the way, I plan on releasing another beta release soon. There were some major changes in the Scene2D modules (#268, #269) - I'd appreciate it if you tried them out.

Quillraven:
got it, thanks! I already used some of the changes in the small new project using the new `scene2d.` syntax. However, it is a small game and does not use it that extensively as the other project. If I have time, I will update the other project to the newest stable version.

czyzby:
@Quillraven Any comments on this particular change? Do you find it useful?

From KTX maintenance point of view, this eliminates code duplication in the module. As for the users, you can finally create all actors with the same DSL and code completion is more accurate (before IntelliJ could auto-complete both extension methods and top-level functions, which cluttered the pop-up and could lead to errors if the incorrect method was used).

Quillraven:
@czyzby from what I can tell it now works better with autocompletion and is cleaner to read. I remember that before I sometimes struggled if I am now using the correct methods and if I am within the ktx DSL or within normal LibGDX parts.

I mentioned you in a PR of the small project with some feedback and looking forward to your input :)

czyzby:
Thanks the input, I hope the changes are a step in the right direction.

I reviewed the PR.

Issue #284, opened 2020-05-18
Add Color destructuring syntax support


heyheyIvan:
Hi!
I suggest adding support for Color destructuring syntax.
It would be great to be able to write code such as:
```kotlin
val (red, green, blue, alpha) = myColor
```

The same is implemented in [androidx](https://developer.android.com/reference/kotlin/androidx/ui/graphics/Color#component1())

czyzby:
Thanks again for the contribution, this seems like a useful utility!

PR #285, opened 2020-05-18
Added support for Color destructuring syntax #284

heyheyIvan:
Its seems like there are enough changes associated with Color class to extract into a separate file.


czyzby:
> It seems like there are enough changes associated with Color class to extract into a separate file.

Agreed.

Thanks for the contribution.

PR #287, opened 2020-06-29
DisposableRegistry

cypherdare:
This is a tool I've been using lately to make it easy to keep track of all the Disposable instances stored in fields of a class so I won't miss disposing of them.

The way I use it is by instantiating a concrete implementation as a delegate. This exposes an `.alsoRegister()` extension function for Disposables so they can be registered right where you instantiate them. Then they are easy to dispose all together. Example:

```kotlin
class MyLoadScreen: Screen, DisposableRegistry by DisposableContainer() {
    private val batch = SpriteBatch().alsoRegister()
    private val logoTexture = Texture("myLogo.png", true).apply {
        setFilter(TextureFilter.Linear, TextureFilter.MipMapLinearLinear)
        alsoRegister()
    }

    //...

    override fun dispose() {
        disposeRegisteredSafely()
    }
}
```

The reason for the interface is to allow you to use it as a delegate so you don't have to subclass it to get the extension functions for Disposable. But subclassing is still an option.

If this sounds like a good fit for Ktx, I can flesh this out with tests and documentation.

### DIFF
--- a/assets/README.md
+++ b/assets/README.md
@@ -80,6 +80,13 @@ collections of assets en masse.
 - All exceptions get a utility `ignore()` method that you can switch at compile time (for debugging or logging) when
 needed. See `Throwable.ignore()` documentation for further details.
 
+#### `DisposableContainer`
+
+- `DisposableContainer` is a `Disposable` that references a set of `Disposable` instances to be disposed all at once.
+When subclassed or used as a delegate via its `DisposableRegistry` interface, it provides an `alsoRegister` extension 
+which allows easily adding items to the container when instantiating them so they'll automatically be disposed when
+the containing class is.
+
 #### `Pool`
 
 - `Pool` instances can be invoked like a function to provide new instances of objects. Basically, this syntax: `pool()`
@@ -155,6 +162,33 @@ val textures: Array<Texture> = getMyTextures() // Works with any Iterable, too!
 textures.dispose() // Throws exceptions.
 textures.disposeSafely() // Ignores exceptions.
 textures.dispose { exception -> } // Allows to handle exceptions.
+```
+
+Registering `Disposable`s for disposal when the containing class is disposed:
+```Kotlin
+import ktx.assets.*
+
+class MyScreen: Screen, DisposableRegistry by DisposableContainer() {
+  val assetManager = AssetManager().alsoRegister()
+  val spriteBatch = SpriteBatch().alsoRegister()
+  // ...
+}
+```
+
+Disposing registered `Disposable`s in a class with its own non-abstract `dispose()` method:
+```Kotlin
+import ktx.assets.*
+
+class MyScreen: ScreenAdapter, DisposableRegistry by DisposableContainer() {
+  val assetManager = AssetManager().alsoRegister()
+  val spriteBatch = SpriteBatch().alsoRegister()
+
+  override fun dispose() {
+    // ScreenAdapter's dispose() hides DisposableRegistry.dispose(), so registry
+    // disposal should be done manually.
+    registeredDisposables.dispose()
+  }
+}
 ```
 
 Scheduling assets loading by an `AssetManager`:
--- a/assets/README.md
+++ b/assets/README.md
@@ -82,7 +82,7 @@ needed. See `Throwable.ignore()` documentation for further details.
 
 #### `DisposableContainer`
 
-- `DisposableContainer` is a `Disposable` that references a set of `Disposable` instances to be disposed all at once.
+- `DisposableContainer` is a `Disposable` that stores a set of `Disposable` instances to be disposed all at once.
 When subclassed or used as a delegate via its `DisposableRegistry` interface, it provides an `alsoRegister` extension 
 which allows easily adding items to the container when instantiating them so they'll automatically be disposed when
 the containing class is.
@@ -175,22 +175,35 @@ class MyScreen: Screen, DisposableRegistry by DisposableContainer() {
 }
 ```
 
-Disposing registered `Disposable`s in a class with its own non-abstract `dispose()` method:
-```Kotlin
+Manually disposing registered `Disposable`s for custom disposal handling or because the superclass hides access to the 
+registry delegate's `dispose` method:
+```
 import ktx.assets.*
 
-class MyScreen: ScreenAdapter, DisposableRegistry by DisposableContainer() {
-  val assetManager = AssetManager().alsoRegister()
-  val spriteBatch = SpriteBatch().alsoRegister()
-
+class MyScreen: ScreenAdapter(), DisposableRegistry by DisposableContainer() {
+  // ...
   override fun dispose() {
-    // ScreenAdapter's dispose() hides DisposableRegistry.dispose(), so registry
-    // disposal should be done manually.
-    registeredDisposables.dispose()
+    registeredDisposables.dispose() { exception ->
+      Gdx.app.error("MyScreen.dispose()", exception.message)
+    }
   }
 }
 ```
 
+Passing a `DisposableContainer` as a primary constructor parameter so its `dispose` method can be accessed in a class
+whose superclass hides the delegate's `dispose` method:
+```
+import ktx.assets.*
+
+class MyScreen(
+  private val registry: DisposableRegistry = DisposableContainer()
+): ScreenAdapter(), DisposableRegistry by registry() {
+  // ...
+  override fun dispose() {
+    registry.dispose()
+  }
+}
+```
 Scheduling assets loading by an `AssetManager`:
 ```Kotlin
 import ktx.assets.*
--- a/assets/src/main/kotlin/ktx/assets/disposables.kt
+++ b/assets/src/main/kotlin/ktx/assets/disposables.kt
@@ -1,6 +1,7 @@
 package ktx.assets
 
 import com.badlogic.gdx.utils.Disposable
+import com.badlogic.gdx.utils.ObjectSet
 import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.InvocationKind
 import kotlin.contracts.contract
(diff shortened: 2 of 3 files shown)


assets/src/main/kotlin/ktx/assets/disposables.kt
```diff
@@ -98,3 +99,97 @@ inline fun <Asset : Disposable> Array<Asset>?.dispose(onError: (Exception) -> Un
...
+
+  override fun disposeRegisteredSafely(deregisterAll: Boolean): Boolean {
+    val hadItems = !registry.isEmpty
+    registry.forEach(Disposable::disposeSafely)
+    if (deregisterAll) registry.clear()
+    return hadItems
+  }
+
+  override fun deregisterAllDisposables(): Boolean {
+    return !registry.isEmpty.also { registry.clear() }
```
czyzby, line 193: This is pretty confusing at first glance, I wouldn't mind this being not a one liner.

czyzby:
It might clash a bit with `AssetManager`, `AssetStorage` and `ktx-inject` `Context`, but I could see it being useful if you want to manage disposables manually. I'd only suggest to simplify the `DisposableRegistry` interface by extending `Disposable` (so you can leverage existing utilities like `disposeSafely`, or nest registries). I think the final API should consist of:

- `dispose()` - inherited from `Disposable`. By default, does not remove the assets.
- `register(Disposable)`
- `deregister(Disposable)`
- `clear()` - removes all registered assets without disposing.

You can also keep/add these optional utilities:

- `val registeredAssets` - a read-only property that returns a copy of the registered assets in a list.
- `Disposable.alsoRegister()`
- `Disposable.alsoDeregister()`

Also, since assets are not guaranteed to implement `hashCode`/`equals` correctly, I'd use an identity set instead of regular set to store the disposables.

cypherdare:
I don't think it clashes with asset management tools because I primarily use it for non-asset Disposables like SpriteBatch, FrameBuffer, programmatically created Meshes, and even AssetManager itself. I don't use dependency injection in my projects.

The reason I named the function `disposeRegistered()` instead of inheriting from Disposable and calling it `dispose()` is this would cause the functionality to be hidden if you use this class as a delegate in something like Screen or especially ScreenAdapter. This could create some "gotchas". If you use it with `Screen`, you would not be able to override `dispose()` to do anything else without removing the ability to dispose of the items in the registry. If you used it in ScreenAdapter, `DisposableRegistry.dispose()` flat out would never be called and would provide no means of access.

Not sure how to avoid this except to have the unique, even if verbose name. Even if we don't inherit from Disposable, if we try to give it a `dispose(deresisterAll: Boolean)` function we lose the ability to have a default value for the parameter, because it could have overload conflicts when the class is used as a delegate.

It would be nice if Kotlin provided a way to instantiate a delegate during construction while also assigning it to a property. 

I like your other suggestions.

czyzby:
I'd still argue that implementing `Disposable` is beneficial, as you can use all existing utilities for disposables and nest registries within each other, i.e. a single registry might aggregate multiple other registries without any extra glue code. Method name clash with `Screen` is not a problem, as the purpose of the method is exactly the same - dispose of all assets managed by the screen. You might want to call `dispose` on a registry manually multiple times for some rare cases, but the same applies to a `Screen`.

cypherdare:
What if you want to do actions other than disposal in `Screen.dispose()`?
That’s how you hook into the lifecycle for when a game is being closed, so
that’s where some LibGDX users may be persisting to files or preferences.

czyzby:
Arguably, this should be handled by `pause`. As a workaround, if you only need to clear the registry, you could access the `registeredAssets` and call `clear`. Is there a particular use case in which you'd like to dispose of the registry, but avoid disposing of a screen?

Adding disposable classes without explicitly implementing `Disposable`, like LibGDX often does, leads to code duplication (e.g. `disposeSafely` in this PR) and inconsistent APIs. I think the ability to leverage `Disposable` utilities and nesting outweighs the potential issues of method name clashes.

czyzby:
By the way:

> I don't think it clashes with asset management tools because I primarily use it for non-asset Disposables like SpriteBatch, FrameBuffer, programmatically created Meshes

All of these can be added manually to a an asset manager/asset storage for easy disposal.

cypherdare:
I don't have a specific example of needing to do that. I just know people sometimes save game state in `dispose()`, though personally I'd use `pause`. The agree it makes sense to inherit from Disposable, but I want to avoid creating gotchas as much as is possible. If you use this class in a ScreenAdapter, it's not obvious that `dispose()` will not be called on the registry. Although this is really a Kotlin gotcha for delegates in general, so maybe we shouldn't worry about it and just document it. But I think `clear()` should be something like `deregisterAll()` since `clear()` is such a common function name and it's easy to avoid the clash.

But, you bring up an interesting point about AssetManager. I always thought AssetManager is exclusively for file-based assets. They are referenced by file name. How would you add a non-file asset to it? It makes me want to pause on this PR and think about whether there's a more elegant solution to do something similar utilizing AssetManager. Maybe an interface for a class with an AssetManager that simply defines a `val assetManager`, and then we could give that extension functions similar to `alsoRegister()` but that adds the Disposable to the manager. Something like 

```
fun <T : Disposable> T.toManager(): T = also { assetManager.addNonFileAsset(this) /* However you do this*/  } 
```

czyzby:
> But I think `clear()` should be something like `deregisterAll()` since `clear()` is such a common function name and it's easy to avoid the clash.

OK, agreed about this one.

> How would you add a non-file asset to it?

I remembered overriding an `addAsset` function for asset manager wrapper while I was implementing `AssetStorage`, but I've checked it right now and it's protected:

```java
protected <T> void addAsset (final String fileName, Class<T> type, T asset)
```

If this function was made public, you could basically store assets with any chosen string, treating `fileName` as a key in hash map rather than an actual file path. Your `toManager` extension method would just need a `String` parameter treated as the asset name.

On the other hand, [`AssetStorage`](https://github.com/libktx/ktx/tree/master/assets-async) is more versatile and provides public `add` method that allows you to register any asset that you want to be managed by the storage and easily accessible via `get`. This would cover your use case with a registry - that's why I was a bit hesitant in the first comments, but I could see this being useful.

czyzby:
If you want to leverage `AssetManager` without extending it, these utilities could expose the protected `addAsset` in a thread-safe manner (notice the package):

```kotlin
package com.badlogic.gdx.assets

inline fun <reified T: Any> AssetManager.add(fileName: String, asset: T) {
  add(fileName, T::class.java, asset)
}

fun <T: Any> AssetManager.add(fileName: String, type: Class<T>, asset: T) {
  synchronized(this) {
    addAsset(fileName, type, asset)
  }
}
```

cypherdare:
So maybe, as an alternative to this new DisposableRegistry, we could have the below interface and extension functions. Not sure of the best way to generate a unique key for each object. I can't think of a determinant way of calculating it so below I just keep a global counter to generate them.

```
private val keySource = AtomicInteger()

interface AssetManagerUser {
    val assetManager: AssetManager

    fun <T: Disposable> T.alsoManage(type: Class<T>): T {
        assetManager.add("${type.name}/${keySource.getAndIncrement()}", type, this)
        return this
    }

    inline fun <reified T : Disposable> T.alsoManage(): T = alsoManage(T::class.java)

    fun <T: Disposable> T.unload()= assetManager.unload(assetManager.getAssetFileName(this))
}
```

And a similar interface could be created for AssetStorage.

Advantages here are that
1. We aren't introducing a feature with overlapping functionality to AssetManager/AssetStorage.
2. This is pretty easy to use without making it a delegate. Just override `assetManager`. You probably have an `assetManager` property anyway which may or may not be part of the primary constructor. So we avoid gotchas with the behavior of delegates.
3. This is lighter weight. We don't have to create a concrete implementation to make it easy to expose `alsoManage()`.
4. The `unload` extension function could be handy to have around even for traditional assets.

czyzby:
You can avoid using `keySource` and `getAssetFileName` by leveraging object hash code instead. Combined with asset class name, it's very unlikely that it would cause any clashes.

There's one major advantage in wrapping `AssetManager`/`AssetStorage` over a custom registry implementation - manager and storage are both thread-safe. We'd only have to test if `addAsset` doesn't break anything internally in `AssetManager`, as it's a protected method after all.

To play the devil's advocate:

> We aren't introducing a feature with overlapping functionality to AssetManager/AssetStorage.

We're also aren't introducing any extra features either, just syntax sugar.

```kotlin
batch.alsoManage()
// vs
assetManager.add("batch", batch)
```

In the second example, you could easily reuse the `batch` outside of the class, extracting it from the manager with the `"batch"` key. The same cannot be said about `alsoManage`, which would have to consume (optional?) string parameter or return the asset key.

> This is lighter weight. We don't have to create a concrete implementation to make it easy to expose alsoManage().

This is definitely easier to implement than the registry, but both manager (fully synchronized) and storage (based on coroutines) are more complex than the original registry implementation. For simple things like your use case, registry would be slightly more performant and possibly easier to use, as you don't have to introduce a manager/storage if you don't want to.

> The unload extension function could be handy to have around even for traditional assets.

As long as they're added to the manager in the first place. If they aren't, that's a runtime exception waiting to happen. When you're forced to manage asset keys/file names manually, that's a bit harder to do. Also, remember than not every asset has to be `Disposable`, but might still need unloading (i.e. big deserialized JSON objects).

If we were to introduce these interfaces, I'd seriously consider adding them as examples in the README/wiki as "recipes" that KTX user can adjust to their use cases. I'm not sure if we want this to be a part of the public API, as it doesn't introduce any crucial features over regular `AssetManager`/`AssetStorage` API.

---

Advantages of the proposed registry over `AssetManager`/`AssetStorage`:

- Allows to easily nest `Disposable` registries without introducing multiple managers/storages.
- You can make arbitrary classes `Disposable`, leveraging `Disposable` utilities.
- Simple implementation, does one job well. Manager and storage _can_ be leveraged as `Disposable` aggregates, but their main function is to load and manage assets.

If you'd still like to, you can finish the registry implementation with adjustments after our discussion. You could even add a second thread-safe implementation of the interface that leverages `AssetManager` to store and manage the assets. Things like `AssetManagerUser` could be added as example of possible alternatives to the registry if you need a less invasive API or a more robust solution (that can also load assets etc.).

cypherdare:
Sounds good. Will do.

czyzby:
@cypherdare You can run `gradle linterIdeSetup` to adjust the formatter in your IntelliJ project and `gradle format` to reformat the sources with `ktlint`.

cypherdare:
Thanks. I missed that in the instructions. Sorry for so many commits, but you can squash while merging, right?

czyzby:
Sure, I can. I'll review the code today or over the weekend.

assets/README.md
```diff
@@ -80,6 +80,13 @@ collections of assets en masse.
 - All exceptions get a utility `ignore()` method that you can switch at compile time (for debugging or logging) when
 needed. See `Throwable.ignore()` documentation for further details.
 
+#### `DisposableContainer`
+
+- `DisposableContainer` is a `Disposable` that references a set of `Disposable` instances to be disposed all at once.
```
czyzby, line 85: `references` -> `groups`? `aggregates`?

assets/README.md
```diff
@@ -157,6 +164,33 @@ textures.disposeSafely() // Ignores exceptions.
...
+```Kotlin
+import ktx.assets.*
+
+class MyScreen: ScreenAdapter, DisposableRegistry by DisposableContainer() {
+  val assetManager = AssetManager().alsoRegister()
+  val spriteBatch = SpriteBatch().alsoRegister()
+
+  override fun dispose() {
+    // ScreenAdapter's dispose() hides DisposableRegistry.dispose(), so registry
+    // disposal should be done manually.
```
czyzby, line 188: That's not entirely true, you can keep a reference to the container and invoke its method:
```kotlin
class DisposableScreen(
  val registry: DisposableRegistry = DisposableContainer()
) : ScreenAdapter(), DisposableRegistry by registry {
  override fun dispose() {
    registry.dispose()
  }
}
```

I would show this in the example and provide both options side by side - manual unloading of all assets and leveraging a registry.

assets/src/main/kotlin/ktx/assets/disposables.kt
```diff
@@ -98,3 +100,85 @@ inline fun <Asset : Disposable> Array<Asset>?.dispose(onError: (Exception) -> Un
...
+
+/**
+ * An implementation of [DisposableRegistry] that can be subclassed or attached as a delegate to
+ * extend DisposableRegistry's convenience functions to [Disposable] declarations within the class.
+ */
+open class DisposableContainer : DisposableRegistry {
+
+  private val registry: MutableSet<Disposable> = Collections.newSetFromMap(IdentityHashMap())
+
+  override val registeredDisposables: Set<Disposable> get() = registry.toSet()
```
czyzby, line 171: Note that this will not create an identity set; it will return a `LinkedHashSet` (afaik) instead. You might need to change this to `.toCollection(Collections.newSetFromMap(IdentityHashMap()))`.

assets/README.md
czyzby: Also, when showing example for collection disposing, you can use `dispose { error -> }` to give the user an idea why manual disposing can be beneficial (i.e. custom error handling).

assets/README.md
cypherdare: Maybe collects?

czyzby: `stores` might make sense.

assets/src/main/kotlin/ktx/assets/disposables.kt
cypherdare: I researched this some more, and the way I did it should be fine. I wrote a test for it, and it passes that part, but `registeredDisposables` fails because `toSet()` doesn't give you an identity `Set`. I opted to change the return type to `List`, since the concept of an identity `Set` technically violates the `Set` contract.

We could alternatively use GDX's IdentityMap directly and just work with its keys since the field is private and we only interact with it minimally. Not sure if it would perform slightly better than the IdentityHashMap. But it's not likely to be frequently accessed in the game loop anyway, so maybe there's no point in adding that complexity.

Issue #14, opened 2016-07-31
KTX tools module


czyzby:
An equivalent to `gdx-tools` that could automate some KTX-related operations. Potential candidates:
- Conversion of `.properties` files to type-safe i18n enums from `ktx-i18n`.
- Conversion of skin `.json` definition to type-safe builders from `ktx-style`.
- Generating type-safe `ktx-style` builders from an existing `Skin` instance.
- Extraction of texture region names from `.atlas` files, which could be saved as a type-safe enum. Useful for both `ktx-style` builders and general `TextureAtlas` usage.

Could be implemented as standalone library or Gradle plugin. Or both. The first approach seems the most flexible. 

@kotcrab

sreich:
fyi so there's no duplication, consider me working on the "Extraction of texture region names from .atlas files, which could be saved as a type-safe enum. ". 

i'm currently waiting on someone else's very recent work, for me to generate the code..i'll see what i can whip up for my own project and go from there. it should be able to generate type-safe enums from all asset types (fonts/images/atlas etc). i'm sick of plain strings :)

czyzby:
That's great! I'll create a separate branch and setup Gradle submodule for you.

czyzby:
There. Try checking out `feature/ktx-tools` branch.

sreich:
think i'm running into an issue with the gradle scripts for ktx (in general, unrelated to ktx-tools). 

resources aren't getting stored in the jar..i'm thinking this is because the top-level one ends up overriding the default behavior (gradle defaults to copy the resources as expected).

you can try adding a src/main/resources/blah.txt to anything and see it not get copied. looks like the resourceDirs get wiped somehow?

czyzby:
@sreich How are you building the jar? I added `src/main/resources` folder in one of the modules, put a text file there and it was present in the generated jar when using `gradle build` command. Maybe this is specific to Gradle plugin jars - have you checked if they are allowed to have classpath resources? I don't see a reason why they wouldn't, but still.

sreich:
guess it's been a long day, silly mistake...i was putting them into src/resources, and since the resource is how gradle finds the plugin in the first place..that is why it couldn't find it. now the plugin works great. so i'll continue working on it and the generator

czyzby:
It's OK. Might I ask why do you need classpath resources in the plugin in the first place?

sreich:
from my understanding of gradle, it is required for the plugin to get picked up by gradle (the implementation-class)

[...]

PR #290, opened 2020-07-12
Tools module and BundleLineCreator for #14

cypherdare:
This adds a class for generating BundleLines mostly automatically. It can be used from an app or the module can be imported as a `buildscript` dependency and used as a gradle plugin that adds a task for generating the BundleLines. See the README for a complete description.

How rigorous should the testing be? I'm not sure how to appropriately test some of the behaviors like finding the assets directory. Unlike other Ktx utilities, this tool is more like a development script, not a component relied on by game code.

Feel free to edit this branch. I don't know if I've appropriately set up the new module.


czyzby:
Thanks for this contribution! I originally wanted to make a Gradle plugin with some KTX utilities, but never went far. See #14.

> How rigorous should the testing be?

Testing the generated output (as a string) would be a good start. I agree that we don't need as rigorous testing as in case of the libraries, but I'd still like the code to be verified automatically.

> I'm not sure how to appropriately test some of the behaviors like finding the assets directory.

If difficult to setup otherwise, write a function that searches for properties files in any set of directories (e.g. temp folders in case of tests) and assume it works correctly given the LibGDX default asset directories.

> I don't know if I've appropriately set up the new module.

I believe it's OK. You can go through [this contribution section](https://github.com/libktx/ktx/blob/master/.github/CONTRIBUTING.md#adding-a-new-ktx-module) to see if you missed anything.

---

Just a heads up - I'm going on a 1 week vacation without internet access. I'll review the PR after I'm back.

cypherdare:
OK. I've added some tests. One piece I don't have a practical test for is checking the validity of the generated code itself.

cypherdare:
I also refactored a bit to better encapsulate the tool and set a pattern for implementing additional tools in the future that can both be used as Gradle tasks or from a JVM app. Here's an explanation of the setup.

In the spirit of many LibGDX classes, the BundleLinesCreator class is open, and most of its methods are open. This allows someone to subclass it and modify some of the behaviors, such as how it selects which files to process. But it has a companion object with the default behavior, so it can be used directly with `BundleLinesCreator.execute()`. Following the precedent from the stdlib `Random` class, the companion object is named `Default`.

The execute function also has an overload where you can pass a single parameters object. This is `internal` because it's only intended for the Gradle plugin to use. The Plugin only has to call `BundleLinesCreator.execture()` and pass it the parameters from the `KtxToolsPluginExtension` object. The parameters object helps organize settings in the plugin extension.

So, for any future utility classes, another parameters object property can be added for it in `KtxToolsPluginExtension`. This will allow all the task parameters to be set in one block:

```groovy
ktxtools {
    createBundleLines.targetPackage = "com.game"
    someOtherTask.option = true
}
```

cypherdare:
I added a test to check the actual produced code. It's not an exhaustive test for everything that could matter as that would be a mammoth task to undertake. It uses regex to extract various elements that we can check easily. There are pieces that aren't useful to check because it would amount to copy-pasting pieces of code we want to generate into the test. For the pieces of code that are checked, the test ignores formatting and syntax variations as long as they're valid (such as type of whitespace, optional newlines, and optional whitespace).

The generated code matches what the default Kotlin formatter in IntelliJ does, as this is the most likely desired formatting.

tools/README.md
```diff
@@ -0,0 +1,75 @@
+# KTX: Tools
+
+Utilities for working with Ktx or generating code for use with LibGDX.
```
czyzby, line 3: We capitalize all letters in KTX.

tools/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Why?
+
+Like the LibGDX `gdx-tools` library, this library contains tools for development rather than use in your game itself.
```
czyzby, line 7: I think it's worth linking `gdx-tools` somehow, but I'm not sure if it's fully documented anywhere. There are a couple of related pages on the official wiki (related to tools such as texture packer), but nothing that would give an overview of the library. Maybe [the repository](https://github.com/libgdx/libgdx/tree/master/extensions/gdx-tools)?

tools/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Guide
+
+This module contains tools that should be not be included in the core module. It can be imported as a `buildscript`
```
czyzby, line 11: Modify to something along:

> (...) should not be directly included in any application modules, including `core`. Instead, it should be imported (...)

tools/README.md
```diff
@@ -0,0 +1,75 @@
...
+### Guide
+
+This module contains tools that should be not be included in the core module. It can be imported as a `buildscript` 
+`classpath` dependency so its Gradle plugin can be used. The plugin adds Gradle tasks to the project in the `ktx tools` 
+group. It can be used by adding this to the `build.gradle` file:
+
+    apply plugin: "io.github.libktx.tools"
+    
+The tools can also be used from an application. In this case, import this library as a dependency of a JVM module (not 
+the `core` game module.)
```
czyzby, line 18: This is a bit confusing, I'd rephrase it to emphasize that it's a separate application:

> The KTX tools can also be used within scripts, libraries or other separate development applications. In this case, you can import this library as a regular dependency of a JVM module. However, make sure that this library is not included in your game modules (such as `core`), as it will unnecessarily increase the application size.

tools/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+#### BundleLineCreator
+
+`BundleLineCreator` is used to generate `BundleLine` enum classes from properties files for use with `ktx-i18n`. It
```
czyzby, line 22: Link to `../i18n`.

```
[`ktx-i18n`](../i18n)
```

tools/README.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+and then running the `createBundleLines` task in the `ktx tools` group. IntelliJ's 
+[FileWatchers](https://www.jetbrains.com/help/idea/using-file-watchers.html) plugin may be of interest as it can be set 
+up to run the task when a file changes.
+
+See usage examples below for how to customize the behavior. Note that `BundleLineCreator` can also be subclassed to 
+customize its behavior farther.
+
+##
```
czyzby, line 46: Remove.

czyzby: If you wanted to add a line separator, use `---` instead.

tools/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Usage examples
+
+Using all the customization options when running `BundleLinesCreator`.
+
+In a JVM app:
```
czyzby, line 52: Change the order of the usage examples, as users are more likely to use the Gradle script and it should be encouraged.

tools/build.gradle
```diff
@@ -0,0 +1,17 @@
+
+plugins {
+  id 'java-gradle-plugin'
+}
+
+dependencies {
+  implementation "commons-io:commons-io:2.4"
+}
+
+gradlePlugin {
```
czyzby, line 10: Is there anything else that we have to add in order to make this plugin available for the users? I haven't published any plugins yet and I'm wondering if I have to add any new Maven repos in order to make it work.

tools/gradle.properties
```diff
@@ -0,0 +1,2 @@
+projectName=ktx-tools
+projectDesc=Development tools for LibKtx
```
czyzby, line 2: `LibKtx` -> `LibKTX.`

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
```diff
@@ -0,0 +1,257 @@
...
+    codeIndent: String = "    ",
+    pathPrefix: String = ""
+  ) {
+    require(codeIndent.all(Char::isWhitespace)) { "codeIntent must consist entirely of whitespace." }
+    val parentDirectory = explicitBundlesDirectory?.let { File(pathPrefix + it) }
+      ?: findFallbackAssetDirectory(pathPrefix)
+    requireNotNull(parentDirectory) { "Failed to find asset directory." }
+    val baseFiles = findBasePropertiesFiles(parentDirectory, searchSubdirectories)
+    val enumNamesToEntryNames = collectEnums(baseFiles, enumClassName)
+    val outDir = File("$pathPrefix$targetSourceDirectory/${targetPackage.replace(".", "/")}")
```
czyzby, line 61: Does it automatically use the correct file separator character? I imagine this wouldn't work on Windows.

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
```diff
@@ -0,0 +1,257 @@
...
+      .apply { mkdirs() }
+    for ((enumName, entryNames) in enumNamesToEntryNames) {
+      val outFile = File(outDir, "$enumName.kt")
+      val sourceCode = generateKtFileContent(targetPackage, enumName, entryNames, codeIndent)
+      outFile.writeText(sourceCode)
+    }
+    println(
```
czyzby, line 68: I don't think this is the correct way of logging, we should rely on Gradle logger instead so it honors the logging level settings.

https://docs.gradle.org/current/userguide/logging.html

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
```diff
@@ -0,0 +1,257 @@
...
+
+  /**
+   * Collects properties files that will be processed. The default implementation finds all files with name suffix
+   * `.properties` and no underscore in the name.
+   * @param parentDirectory The directory that will be searched.
+   * @param searchSubdirectories Whether to include sub-directories of the parent directory in the search.
+   * @return A list of all applicable properties files found.
+   */
+  protected open fun findBasePropertiesFiles(parentDirectory: File, searchSubdirectories: Boolean): Collection<File> {
+    val noUnderscoresFilter = object : AbstractFileFilter() {
```
czyzby, line 102: I believe you don't need this library to achieve this on JVM, try using `java.nio.file.Files#walk`.

czyzby: Also, I'd be surprised if Gradle didn't have similar utilities.

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
```diff
@@ -0,0 +1,257 @@
...
+   * @return A map of enum names to the enum entry names that should be members of those enums.
+   * */
+  protected open fun collectEnums(
+    propertiesFiles: Collection<File>,
+    commonEnumName: String?
+  ): Map<String, Set<String>> {
+    val outMap = mutableMapOf<String, MutableSet<String>>()
+    val sanitizedCommonBaseName = commonEnumName?.let { convertToEnumName(it) }
+    if (sanitizedCommonBaseName != commonEnumName)
+      println("The provided enumClassName $commonEnumName was changed to $sanitizedCommonBaseName.")
```
czyzby, line 128: Use a warning log.

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
```diff
@@ -0,0 +1,257 @@
...
+   * @return A name based on [name] that is a valid enum entry name, or null if the input name cannot be used as a valid
+   * enum entry.
+   * */
+  protected open fun convertToEntryName(name: String): String? {
+    if (name.isEmpty()) {
+      println("A blank property name was encountered and will be omitted.")
+      return null
+    }
+    if ('`' in name) {
+      println("The property name \u001B[0;31m${name}\u001B[0m cannot be converted into a usable enum entry and will be omitted.")
```
czyzby, line 168: Do not use UTF codes, change to `'`.

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
```diff
@@ -0,0 +1,257 @@
...
+      append("@Suppress(\"EnumEntryName\")\n")
+      append("enum class $enumClassName : BundleLine {\n")
+      for ((index, key) in entryNames.withIndex()) {
+        val lineEnding = if (index < entryNames.size - 1) "," else ";\n"
+        append("${indent}$key$lineEnding\n")
+      }
+      append("${indent}override val bundle: I18NBundle\n")
+      append("${indent}${indent}get() = i18nBundle\n\n")
+      append("${indent}companion object {\n")
+      append("${indent}$indent/** The bundle used for [BundleLine.nls] and [BundleLine.invoke] for this enum's values. */\n")
```
czyzby, line 205: Add to documentation:

> Must be set explicitly before extracting the translated texts.

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
```diff
@@ -0,0 +1,257 @@
...
+}
+
+/**
+ * Parameters for running [BundleLinesCreator], intended for the `createBundleLines` Gradle task.
+ * */
+open class BundleLinesCreatorParams {
+  /**
+   * The package created enums will be placed in. Must be set, or the `createBundleLines` task cannot be used.
+   * */
+  lateinit var targetPackage: String
```
czyzby, line 220: Just to confirm, you're using an `lateinit var` because it is automatically constructed by Gradle?

tools/src/main/kotlin/ktx/tools/KtxToolsPlugin.kt
```diff
@@ -0,0 +1,33 @@
+@file:Suppress("unused")
+
+package ktx.tools
+
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+
+internal const val TASK_GROUP = "ktx tools"
+internal const val PROJECT_EXTENSION = "ktxTools"
```
czyzby, line 9: Maybe we should just go with `"ktx"`?

tools/src/test/kotlin/ktx/tools/BundleLinesCreatorTest.kt
```diff
@@ -0,0 +1,210 @@
...
+      checkBrackets('(', ')')
+    )
+    assertTrue(
+      "Improperly matched curly braces in generated code:\n$code",
+      checkBrackets('{', '}')
+    )
+    assertTrue(
+      "Improperly matched square brackets in generated code:\n$code",
+      checkBrackets('[', ']')
+    )
```
czyzby, line 191: I think it makes sense to just compare it with an equality check to a multiline string with the expected output. Checking the matching brackets says nothing about the validity of the code.

czyzby:
> There are pieces that aren't useful to check because it would amount to copy-pasting pieces of code we want to generate into the test.

That's actually not a bad way to test this, in my opinion. After all, in typical tests you'd ideally like to end up with a single equality check of the produced output, not a bunch partial checks verifying only parts of the result. I think it's reasonable to copy or write the expected code in a multiline string and compare it against the generated class, as it would be a better indicator of valid code generation than e.g. matching closing brackets - and it's easy to add extra tests for settings like custom class name or indent.

Thanks for the contribution, looking good so far.

tools/README.md
cypherdare: Remove the whole paragraph or just the last sentence here?

tools/build.gradle
cypherdare: I'm not sure, as I don't have a lot of experience setting this up on the publishing side. I did test this by running the `install` task and then using it in one of my game projects via `mavenLocal()`. I don't know if there it anything else you need to check to make sure it is published automatically along with everything else because I haven't familiarized myself with the various scripts you have set up for publishing.

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
cypherdare: Java converts these to backslashes under the hood on Windows, but it is apparently [not really documented clearly](https://stackoverflow.com/a/2417546/506796). I'll just use `File.separator` to do it properly.

tools/build.gradle
czyzby: Alright, I guess we'll figure it out once it's merged.

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
cypherdare: Yes. I suppose it would be less weird to make it nullable with default value of null, but call `requireNotNull` on it from the `execute` method. I think I'll do that unless you like this better.

tools/src/main/kotlin/ktx/tools/KtxToolsPlugin.kt
cypherdare: For both of these, right?

czyzby: Sure.

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
czyzby: Either way works for me.

tools/src/test/kotlin/ktx/tools/BundleLinesCreatorTest.kt
cypherdare: OK, replace the entire test with a simple equality check against a single hard-coded String? I wasn't sure about that because there are changes to the output we could make that wouldn't be errors, but I suppose we could just update the test accordingly if we did that.

tools/README.md
cypherdare: Got it. I misunderstood at first. That was just a copy-paste remnant.

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
cypherdare: The ktx project as a whole is set to Java compatibility 1.6, right? Not sure if we want to have to set up this module separately to be able to use Java 8 APIs. AFAIK, using the Gradle utilities would eliminate the ability to run this as anything but a Gradle task, as the file walking methods rely on a Gradle Project instance.

tools/src/test/kotlin/ktx/tools/BundleLinesCreatorTest.kt
czyzby: If the output changes and the tests do not catch that, I'd argue that the tests are incomplete. We're supposed to ensure reproducibility and stability of the generated code.

tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt
czyzby: Set this module to 8 explicitly, it's OK - Gradle runs on 8+ either way.

cypherdare:
OK, I got everything updated and resolved. I've tested adding it as a buildscript dependency and running the task, and it runs fine.

But when I add it as a dependency and try to run the `BundleLinesCreator` in an application, it now fails because the Gradle API is missing as a dependency. This is because the `BundleLinesCreator` class now references the Gradle API Logger class.

I think maybe this is because I didn't make the API a dependency in `build.gradle`, but rather used the `java-gradle-plugin`, which automatically adds it as a dependency in the project when I'm working on it. I'm not sure what the distinction is, but I guess by default it doesn't make the API a dependency because it's intended for making plugins, where the Gradle API will obviously always be available.

I haven't looked yet if there is a Gradle API Maven repository I can simply add as an `implementation` dependency  in the `build.gradle` file so this will work properly.

Got kind of burned out on this the past couple days. I will probably take a week or two before I get back to it, but you're welcome to try solving the above problem in the meantime.

czyzby:
> But when I add it as a dependency and try to run the `BundleLinesCreator` in an application, it now fails because the Gradle API is missing as a dependency. This is because the `BundleLinesCreator` class now references the Gradle API Logger class.

Have you tried using SLF4J or any other logger wrapper? I don't think we should spend too much time on this, SLF4J/Log4J do not work with Gradle, we can revert to prints.

> Got kind of burned out on this the past couple days. I will probably take a week or two before I get back to it, but you're welcome to try solving the above problem in the meantime.

Understandable. Thanks again for the contribution. I'll probably wait for you to come back, but I'll review the changes in the meantime.

cypherdare:
I just switched to use SLF4J and it works fine now. One possible issue is that Gradle's logger defaults to "Lifecycle" level which is between SLF4J's "Info" and "Warn". I log the results of the file creation at the info level, so this message won't show when running the task without modifying the log level. This is probably fine, because you don't need this kind of info unless you're trying to debug why the task isn't giving you your expected output.

cypherdare:
Are the tests failing because SLF4J emits a warning about duplicate bindings? I don't know the proper way to resolve this.

czyzby:
> This is probably fine, because you don't need this kind of info unless you're trying to debug why the task isn't giving you your expected output.

That's the default behavior of Gradle. That's OK.

> Are the tests failing because SLF4J emits a warning about duplicate bindings? I don't know the proper way to resolve this.

I don't think so, I believe it's just a warning. If you dig deeper that's the main cause:

```
Execution failed for task ':tools:lint'.
```

My guess is you just have to reformat the sources with `gradle format`.

cypherdare:
Haha, the imports were not alphabetized!

czyzby:
Thanks @cypherdare, looks good now. Could you update to the latest develop changes with LibGDX 1.9.11 and add a changelog entry briefly describing the new module features?

deveth0:
@cypherdare awesome work, thank you so much! I tried to integrate the code into a small sample project and ran into some issues regarding the enum names. 
Look at the following file `foo.properties`:
```properties
foo=bar
foo.lorem=ipsum
```

This will generate this enum:
```kotlin
@Suppress("EnumEntryName")
enum class Menu : BundleLine {
    foo,
    `foo.lorem`;
...
```

Sadly this is not valid code. What do you think about changing the generation so that it generates something like this;
```kotlin
@Suppress("EnumEntryName")
enum class Menu : BundleLine {
    FOO("foo"),
    FOO_LOREM("foo.lorem");
...
```

czyzby:
That's actually a good point, we should add a line like that to the test suite.

@deveth0, what would you say about creating a nested enum class, so that the generated code closely resembles the properties file structure? For example:

```kotlin
enum class Menu : BundleLine {
  // ...

  enum class foo {
    forem
  }
}

fun usageExample() {
  Menu.foo.forem
}
```

@cypherdare I'm sorry about dragging this PR for so long, I'll try to merge it this month. If you don't have more time to work on this, I'll change the generation code to support nested enums or better escaping.

deveth0:
Hm, we had a similar discussion in another context and the general agreement was, that property files don't really support nesting and this is a weird hack ;)
I also think that this might result in a real clutter if the "nesting" becomes too deep. 

Actually I was a bit bored and played around with the second part of the implementation, the generation of Asset Atlas enums. You might run into a similar problem there cause the name of the Region is not necessarily a valid Enum name. So I think the safest way is to keep the original String available and use a safe version as identifier. 

Please have a look into this repository:
https://github.com/deveth0/ktx-tools-demo

I stripped down @cypherdare 's code a bit and extracted a generator for Enums using kotlinpoet. The demo-project automatically runs the two tasks (`createBundleLines` and `createAssetEnums`) and places the generated sources in `/build/generated-sources`. 

Imho there is no real reason why those files should not be generated during the build, this also makes sure, that they are always up to date. As they are automatically generated code, I'd also suggest to exclude them from ktlint.

czyzby:
Makes sense, it's similar to how other Java projects work. I pretty sure you can also automatically mark folders with the generated files as source folders when applying the plugin. Kotlinpoet is a nice touch. This is actually a very nice expansion of the original plugin - would you like to submit your code as a PR? I could setup a separate branch for `ktx-tools` with @cypherdare's changes that you can fork.

cypherdare:
Good catch.

I don't have time in the very near future to add nesting, so if either of you wants to flesh this out more, that's fine with me. 

In my opinion, it should not automatically change property names into SNAKE_CASE. I personally prefer PascalCase for enum constants. As it is, you can control the case type yourself to your preference by your choice in the properties file.

I originally thought about using KotlinPoet, but thought it was overkill originally. But it isn't if you're adding nesting.

What is the reason to break out the EnumCreator abstract class?

czyzby:
I agree about preserving the case. While it might require adjustments in case of the texture atlases, I think it's the most flexible approach for i18n bundles - we're simply letting the users decide how their identifies look.

As for the nesting - I don't necessarily see this as a hack, as it allows to preserve the user-defined structures within the properties files (even if somewhat artificial). It also has the additional benefits of supporting selective imports - e.g. you can star-import only specific bundle lines assigned to a chosen screen. I could see that as better than just generating a single enormous enum and basically forcing the users to split the bundle into multiple files to achieve a similar (but less convenient) effect.

The only problem that I can see is the issue with name collisions. For example, given `foo` and `foo.bar` properties, you cannot both add a `foo` bundle line and a `foo` enum class to support nesting. To fix that, I'd add a plugin setting that allows the user to either issue a warning or throw an error during the generation when encountering.

And by the way, changing the properties' names to uppercase wouldn't entirely solve the problem of name collisions either. For example, if we replaced special characters with underscores, both `foo.bar` and `foo-bar` would end with the same `FOO_BAR` identifier. It also loses the benefit of grouping chosen properties in code. If both approaches end up with different issues, I'd rather go with the one that's more flexible and the one that generates code closer to the original users' configuration.

deveth0:
Thank you both for your feedback, really appreciate that!
@cypherdare I needed a plugin for my AssetAtlasEnum and as a lazy developer, I was looking for some inspiration and found your code. So the natural choice was to analyze, which parts I could reuse and this resulted in the `EnumCreator`. 
This was also the main reason why I replaced the static generation with `kotlinpoet`, it's more flexible than manipulating strings in subclasses. 

So both of you agree, that the enum type names should keep the original names of i18n keys and use nesting? Fine with that, the `*` import is a rather good argument!

@czyzby as I have virtually NO clue, how to use a gradle plugin from another project, I'd prefer to keep the code in the demo project I setup and play around a bit with nesting etc until we come to a state where we all agree, that this is the way to go. I'd then take the ktx-tools branch and integrate my changes.

czyzby:
@deveth0 Sure, take your time.

deveth0:
Looking at the code again: Is there really a reason, why multiple files should be merged into the same Enum?
This brings in some potential problems we can easily avoid by removing this option. Mainly duplicates become hard to detect as the second entry just overrides the first one. Also it's no longer possible to track the origin of the entry. 

To be honest I never used the i18n feature in libgdx so I might just be ignorant here ;)

czyzby:
If the `.properties` files have the same core name (as in, they are different variants/languages of the same properties set), we should gather all properties and put them in the same enum. It's essentially the same translation bundle.

As for multiple properties files with different names, I think generating separate enums might be even easier to manage for the users code-wise, as you might need to load them with separate `I18NBundle` instances either way.

czyzby:
If you want to work on this, I think it might be a good idea to give `ktx-i18n` a try just to see how it works.

deveth0:
@czyzby I definitly will ;) Merging all values into a single enum is a feature @cypherdare has implemented, I'm just not sure, if this is required.

czyzby:
OK, so to clear this up - given these files:
- `MyBundle.properties`
- `MyBundle_en.properties`
- `MyBundle_fr.properties`
- `OtherBundle.properties`

The plugin should generate two enums: `MyBundle` with properties from the first three files with the same core name and `OtherBundle` with properties from the last file.

deveth0:
Exactly :)  I was refering to https://github.com/cypherdare/ktx/blob/tools/tools/src/main/kotlin/ktx/tools/BundleLinesCreator.kt#L259

czyzby:
Right, so this is something we can change before merging the `ktx-tools` into `master`. I think it makes more sense to separate different bundles into multiple files, even from the practical point of view - the users might want or need to maintain multiple separate bundles, and putting all properties in a single enum only makes it more error-prone. The plugin property that you highlighted can be removed and the class name can be extracted from the file name. Let's assume that anything before the first underscore is the core name of the bundle.

czyzby:
I'm going to close this PR for now, as the plugin currently does not generate multiple separate bundles correctly. If either of you would like to work on this one @cypherdare @deveth0, feel free to set up another PR. Thanks.

PR #311, opened 2020-10-25
Updating gradle to latest + kts

Megamiun:


actors/build.gradle.kts
```diff
@@ -0,0 +1,6 @@
+val gdxVersion: String by rootProject
```
czyzby, line 1: Is there a way of making it global or accessing it without the variable redefined over and over?

czyzby:
I believe some of the tests don't pass because the modules that require one another (e.g. `assets-async`) are not properly linked.

czyzby:
@Megamiun Merges to `master` are only for releases. The actively developed branch is `develop`. Please change the target branch.

actors/build.gradle.kts
Megamiun: Hello, unfortunately I didn't find any way to apply it to plugins, but I did get a way for the rest of them!

czyzby: Alright, thanks.

Megamiun:
> @Megamiun Merges to `master` are only for releases. The actively developed branch is `develop`. Please change the target branch.

@czyzby I read your message wrong and rebased to develop too, should I go back?

czyzby:
@Megamiun I'll squash the commits, it's OK. As long as the target branch is fine and the build passes, it should be alright.

build.gradle.kts
```diff
@@ -0,0 +1,280 @@
...
+      jvmTarget = "1.6"
+      freeCompilerArgs += "-Xopt-in=kotlin.RequiresOptIn"
+    }
+  }
+
+  dependencies {
+    val kotlinVersion: String by project
+
+    "provided"("org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion")
+    "provided"("com.badlogicgames.gdx:gdx:$gdxVersion")
```
czyzby, line 88: Pretty sure `nebula.provided-base` plugin is no longer required and you should use `compileOnly` in newer Gradle versions. Could you test if a similar POM is generated with `compileOnly`?

czyzby (commented):
Looking very clean so far, I can see the improvement over Groovy. We could probably merge this as is, but I'd appreciate it if you could look into the provided plugin that I've mentioned.

build.gradle.kts
Megamiun: Ok, I will take a look at that!

Megamiun: Had to make some extra changes and discovered that the pom was not being correctly created D:

But know everything seems to be working as expected

czyzby (approved):
This is looking very promising, I'll be happy to test this setup during the nearest release. I'll merge this as is and we'll see if the snapshot publication task can be performed successfully.

Megamiun:
> This is looking very promising, I'll be happy to test this setup during the nearest release. I'll merge this as is and we'll see if the snapshot publication task can be performed successfully.

Happy to help :D Anything more you need related to this upgrade, tell me.

Oh, and also, I think I saw an issue about Dokka, is there any interest of me trying to update it?

czyzby:
@Megamiun I've tried migrating to the latest Dokka version, but it failed with out-of-memory errors - at least when we were still on Gradle 5. I don't think I've kept the branch, but the update itself should be pretty simple. However, if it's not related to Gradle, chances are it would also fail in our CI environment, which likely has less dedicated resources than my local setup. Latest Dokka brings some interesting features like the support for multi-module projects, so if you'd like to work on this, you're welcome to try. Maybe they fixed a couple bugs since then. :)

czyzby:
@Megamiun There's actually a small problem with [the uploaded snapshots](https://oss.sonatype.org/content/repositories/snapshots/io/github/libktx/). Archive names do not feature the `ktx-` suffix, so for example `ktx-assets` became just `assets`. It should be an easy fix, I'll work on this right now.

czyzby:
Turns out the Maven configuration was missing `artifactId`. It seems that you only needed to set the `name` previously, but now `artifactId` defaulted to Gradle module name.

czyzby:
@Megamiun There were some changes in the publishing plugin that prevent me from uploading the library to Maven Central. `uploadArchives` no longer publishes all artifacts to Maven Central. It just immediately passes. I've tried `publishAllPublicationsToMavenRepository` instead, but it doesn't seem to publish sources or javadoc archives, so the repository does not pass the validation in Maven Central. I'll try to fix this locally, but if you have any idea what might have been changed in the process and how to fix this, let me know.

czyzby:
Not sure if this is still relevant, but the `maven-publish` [doesn't seem to support](https://discuss.gradle.org/t/how-to-publish-artifacts-signatures-asc-files-using-maven-publish-plugin/7422/30) uploading signed filed out of the box.

czyzby:
Adding:

```kotlin
          artifact(sourcesJar)
          artifact(javadocJar)
```

To the Maven publication setup uploads the javadoc and sources jars, but I think keys (asc files) are still not uploaded. I'll try to fix it later, but if you have any ideas how to solve this in Kotlin DSL, I'd appreciate a PR.

czyzby:
Just wanted to let you know that I've got it, there was a slight change that you had to do around the signing configuration.

PR #316, opened 2020-10-31
Updating Dokka to 1.4.10.2

Megamiun:


czyzby:
Thanks for implementing this so fast, I'll try to review and test this tomorrow.

Megamiun:
@czyzby No problem, man, it's a pleasure :D

And now it seems to be working ok on both my computer and on Github Actions now. I had some OutOfMemory on my computer too before adding the property.

The only thing I didn't get to port between the versions was the disclaimer that existed previously on the index, I don't know if is a problem or not

czyzby:
It's not. As long as the documentation for all modules is combined, it's OK.

.github/workflows/publish-documentation.yml
```diff
@@ -18,9 +18,7 @@ jobs:
       - name: Execution permission for Gradle wrapper
         run: chmod +x gradlew
       - name: Dokka
-        run: ./gradlew dokka
-      - name: Gather modules documentation
-        run: ./gradlew gatherDokkaDocumentation
+        run: ./gradlew dokkaHtmlMultiModule
       - name: Publish GitHub pages
```
czyzby, line 22: I haven't seen the generated output yet, but I'm assuming you might have to modify the folder in this step to `build/dokka/html` judging by the other changes.

Megamiun: Yes, I had to, didn't see this one hahahahhaha

czyzby:
I'm so glad Dokka no longer spams hundreds of unnecessary warnings.

.github/workflows/publish-documentation.yml
```diff
@@ -18,12 +18,10 @@ jobs:
...
         uses: JamesIves/github-pages-deploy-action@releases/v3
         with:
           ACCESS_TOKEN: ${{ secrets.accessToken }}
           BRANCH: gh-pages
-          FOLDER: build/dokka
+          FOLDER: build/dokka/html
```
czyzby, line 27: The built documentation is actually at `build/dokka/htmlMultiModule` and lacks an `index.html` page.

czyzby (changes requested):
Documentation is generated correctly and I didn't run into any memory issues. However, the generated multi-module documentation has two issues:

- It lacks an `index.html` page, which means it would not display properly as-is.
- I haven't found any way to go back to the main page with all modules via the UI.

I think before merging this we should address these issues by:
- Renaming `-modules.html` to `index.html` in the multi-module documentation.
- Replacing references to `-modules.html` with `index.html` (AFAIK there's just one link and it's in the `-modules.html` file itself).
- Adding a link to the main documentation page to the logo (`<div id="logo"></div>`). If it proves too difficult to make a relative path to the root directory with all of the modules, we can link to something like `https://libktx.github.io/docs/`.

Most of these can be achieved via Gradle by rewriting the generated files. Let me know if you want me to take over @Megamiun.

.github/workflows/publish-documentation.yml
```diff
@@ -18,12 +18,10 @@ jobs:
       - name: Execution permission for Gradle wrapper
         run: chmod +x gradlew
       - name: Dokka
-        run: ./gradlew dokka
-      - name: Gather modules documentation
-        run: ./gradlew gatherDokkaDocumentation
+        run: ./gradlew dokkaHtmlMultiModule
```
czyzby, line 21: I've looked into it and we might actually prefer `dokkaHtmlCollector`, as it builds the documentation as if it was a single module. Makes the search and navigation much easier, but it still does not generate the top level `index.html` file for some reason.

.github/workflows/publish-documentation.yml
czyzby: Since we'll be moving to the `dokkaHtmlCollector` task, the folder will be `build/dokka/htmlCollector`.

Megamiun:
Hello, @czyzby, I moved the kts/ index to the root and changed the navigation to point back. It isn't perfect, but I think it causes less problems than moving the entire structure, in wich case, I would have to apply a regex to all files.

Megamiun:
On reality,  I think we can maintain the navigation file as is and have a /index and a /kts/index too.

This way the person arrives at /index and all subsequent navigations use kts/index and /kts/module/index.

Your choice there. But if this is desired, just cleaning the navigation rewrite I did will do the job.

czyzby:
This is OK, but AFAIK there's still an issue with some links to submodules. I've hosted it on a local HTTP server and ran into 404 as the main `index.html` was pointing to e.g. `ktx.actors/` rather than `ktx/ktx.actors/`.

I think it might be easier to just make a redirection page rather than to duplicate. Try adding a similar root level `index.html` after the Dokka task:

```html
<!DOCTYPE HTML>
<html lang="en-US">
    <head>
        <meta charset="UTF-8">
        <meta http-equiv="refresh" content="0; url=ktx/">
        <script type="text/javascript">
            window.location.href = "ktx/"
        </script>
        <title>KTX documentation</title>
    </head>
    <body>
        <!-- Note: don't tell people to `click` the link, just tell them that it is a link. -->
        If you are not redirected automatically, follow <a href="ktx">this link</a>.
    </body>
</html>
```

czyzby:
I might move the documentation to a separate repository like `docs`, so the URL becomes `docs/ktx` rather than `ktx/ktx`.

Megamiun:
Ops, just one moment, I think there is still a problem

build.gradle.kts
```diff
@@ -237,46 +238,42 @@ subprojects {
...
+            <head>
+                <meta charset="UTF-8">
+                <meta http-equiv="refresh" content="0; url=ktx/index.html">
+                <script type="text/javascript">
+                    window.location.href = "ktx/index.html"
+                </script>
+                <title>KTX documentation</title>
+            </head>
+            <body>
+                If you are not redirected automatically, follow <a href="ktx/index.html">this link</a>.
```
czyzby, line 264: It's actually OK to link to just `ktx/` as long as the documentation is published via a HTTP server, but I'll merge this as is.

czyzby (approved):

czyzby:
I'll update the GH pages deployment branch and repository myself, I've set up a `libktx/docs` repo for that.

Megamiun:
@czyzby Wait just a minute, there is some unnecessary code and some minor improvements yet to do

czyzby:
What's the issue? I think this should be ready for a merge.

Megamiun:
@czyzby The dokkaHtmlAddIndex doesn't need the copy operation anymore. And I just tried and didn't saw the navigation bar

czyzby:
Oh, I see. Yeah, all you need to do is create the extra `index.html` right now.

Megamiun:
Ok, @czyzby, I think it is alright for both of us now.

czyzby:
Thanks for another great contribution, @Megamiun.

Megamiun:
Thank you very much too, it was very cool to do those updates, and they were half of my Hacktoberfest contributions too hahahaha

If there is anything more to do about updating things or doing structural changes, you can @ me. I like the idea of the lib, but I am not very good at game logic, so those were the issues I found myself more comfortable solving.

czyzby:
I see. Is there anything I can do to make sure that your contribution is properly counted? I remember something about a special label that you have to assign to avoid being flagged as a spammer.

Megamiun:
Oh, no, it seems that being merged or approved is just enough:

https://hacktoberfest.digitalocean.com/hacktoberfest-update

Megamiun:
As your project already have the hacktoberfest tag, there is no need for further action, it seems

czyzby:
I see. I'll keep the repository label for a while then. Thanks again.

PR #345, opened 2021-02-22
getOrPut extension function for maps (Discussion #341)

Quillraven:
So far I am only missing this function in `ObjectMap` but I added it now also to several other map types.

As always, open for feedback :)


Quillraven:
no idea how to fix the conflict since a pull tells me I am up to date and I can push without a conflict ... In my opinion my version should be the correct one.

czyzby:
It says that: 

> This branch is 5 commits ahead, 7 commits behind libktx:develop.

Are you sure you've pulled the latest changes from `develop` from the `libktx/ktx` repository?

collections/README.md
```diff
@@ -96,6 +96,7 @@ has to be provided - since the method is inlined, no new lambda object will be c
   - `GdxArrayMap`: `com.badlogic.gdx.utils.ArrayMap`
 - All LibGDX map entries now feature `component1()` and `component2()` operator extension methods, so they can be
 destructed into a key and a value.
+- `getOrPut` for `ObjectMap`, `IdentityMap`, `ArrayMap` and `IntMap` method that works like the Kotlin stdlib method
```
czyzby, line 99: Describe the behavior briefly, you don't need to mention stdlib.

collections/src/main/kotlin/ktx/collections/maps.kt
```diff
@@ -422,3 +422,63 @@ inline fun <Key, Type, Value : Iterable<Type>> GdxMap<Key, out Value>.flatten():
...
 }
+
+/**
+ * Returns the value for the given [key]. If the [key] is not found in the map,
+ * calls the [defaultValue] function, puts its result into the map under the given [key] and returns it.
+ */
+inline fun <Key, Value> GdxMap<Key, Value>.getOrPut(key: Key, defaultValue: () -> Value): Value {
+  var value = this[key]
+
+  if (value == null) {
```
czyzby, line 433: This will not work correctly if LibGDX maps can store `null` values. It would be better to check if the key is present in the map. Please add a test case for this.

czyzby: Same goes for other maps.

collections/src/test/kotlin/ktx/collections/MapsTest.kt
```diff
@@ -393,4 +393,82 @@ class MapsTest {
...
     assertEquals("Four", gdxArrayMap[4])
   }
+
+  @Test
+  fun `should return existing value for GdxMap when key exists`() {
+    val map = gdxMapOf("42" to 42)
+
+    val actual = map.getOrPut("42") { 43 }
+
+    assertEquals(42, actual)
```
czyzby, line 403: Ensure that the value was not replaced in the map, i.e. `"42"` should still point to `42`.

czyzby: Same goes for other similar tests.

collections/src/test/kotlin/ktx/collections/MapsTest.kt
```diff
@@ -393,4 +393,82 @@ class MapsTest {
...
+
+  @Test
+  fun `should return and put default value to IntMap when key does not exist`() {
+    val map = IntMap<String>()
+
+    val actual = map.getOrPut(42) { "43" }
+
+    assertEquals("43", actual)
+    assertTrue(42 in map)
+  }
```
czyzby, line 473: Add test cases with `null` values.

czyzby (changes requested):
Try to resolve the conflict in your repo - pull the latest changes from this repository, as it seems that a few commits are missing. Ensure the extensions work as expected given null values.

collections/src/main/kotlin/ktx/collections/maps.kt
Quillraven: The reason I went for that approach is to avoid a second call to `locateKey` which should be slighty more efficient than this approach:

```Kotlin
if(key in map) {
  return map[key] // second time locateKey is called
} else {
  // ...
}
```

But, as always, you are correct that this will not work in case you put null values to the map. I did not consider that and will change it.

czyzby: > The reason I went for that approach is to avoid a second call to l`ocateKey` which should be slighty more efficient

That's what I thought. You can (kind of) avoid the second call with something along the lines of:

```kotlin
val value = map[key]
if (value == null && key !in map) {
  // ...
}
```
In best case scenario, it will perform only a single lookup and return the existing value.

czyzby: Now that I think about, what happens if you try to use a `null` key? It's OK if it throws an NPE error - you don't need to change the function itself, but maybe an extra test for the thrown exception would make sense just to document the behavior.

Quillraven: How can I test that? :D Key is a not nullable type and therefore I cannot pass null?

czyzby: Kotlin type system strikes again, we might not have to test for this after all.

Anyway, you could explicitly define the key type:

```kotlin
val map = gdxMapOf<String?, String?>()
map.put(null, null) // Compiles.
```

Quillraven:
I think I updated the branch now correctly. I pulled from the "upstream" develop branch and got some updates now and also the conflict (which is now resolved).

Also, added now the null test cases in case a map stores null values.

collections/src/main/kotlin/ktx/collections/maps.kt
Quillraven: That is true but does that ever make sense? Why would I use a nullable key in Kotlin?

czyzby: The only reason I can think of is receiving a Java type (`Any!`) and falsely assuming is not null. You can easily avoid this is in pure Kotlin code, but interacting with Java can lead to weird errors like think.

Since type system mostly prevents you from making this error, I'll leave it up to you. If you think the extra tests could be helpful, include them in the PR. Otherwise you can resolve the discussion.

czyzby (approved):
We can merge this as is after all discussions are resolved.

collections/src/main/kotlin/ktx/collections/maps.kt
Quillraven: I added test cases for it and documented the cases where a `java.lang.IllegalArgumentException` is thrown.

`IntMap` does not allow null keys and `GdxArrayMap` somehow supports null keys :D

collections/src/test/kotlin/ktx/collections/MapsTest.kt
```diff
@@ -393,4 +395,154 @@ class MapsTest {
...
+  fun `should return null for GdxMap when null is stored for given key`() {
+    val map = gdxMapOf<String, Int?>("42" to null)
+
+    val actual = map.getOrPut("42") { 43 }
+
+    assertNull(actual)
+    assertEquals(null, map["42"])
+  }
+
+  @Test(expected = IllegalArgumentException::class)
```
czyzby, line 430: In Kotlin we can use `shouldThrow<IllegalArgumentException>` to assert if a certain exception is thrown. In this case it is fine as is, but in general `shouldThrow` is a bit more flexible, as you don't have to end the test with an exception and you can continue with other asserts.

czyzby:
Thanks!

collections/src/test/kotlin/ktx/collections/MapsTest.kt
czyzby: See the latest commit.

Quillraven: Thank you for the hint. I was looking for assertThrows and something similar and then I found that approach in one of the existing test cases in ktx und just reused it.

Will keep that in mind for the future  :)

PR #347, opened 2021-03-03
isEmpty, isNotEmpty and forEachLayer for ktx-tiled

Quillraven:
Some small extensions that I introduced in my own project and that might be useful for others:
- `isEmpty` and `isNotEmpty` for `MapLayers` and `MapObjects` class
- `forEachLayer` for `TiledMap` class. The purpose of this function in my case was that I needed a way to iterate over a specific type of layers. LibGDX has a `getByType` function as part of `MapLayers` class. However, this function internally always creates an array and returns it as a result or you need to pass in an array that gets filled with those layers. 
I did not like that and that is the reason for the `forEachLayer` extension. It does basically the same but without the need of an additional array object.

Let me know if this is useful or not and in case it is useful if it requires changes or not ;)

### DIFF
--- a/tiled/README.md
+++ b/tiled/README.md
@@ -103,11 +103,11 @@ val map: TiledMap = getMap()
 val myFloatProp: Float = mapObj.property("myFloatProperty")
 
 // Retrieves String property with a default value:
-val myProp: String = mapObj.property("myProperty", defaultValue="")
+val myProp: String = mapObj.property("myProperty", defaultValue = "")
 
 // The explicit type can be omitted as it is automatically derived from the type of the default value.
 // myProp2 is of type Float
-val myProp2 = mapLayer.property("myProperty2", defaultValue=1f)
+val myProp2 = mapLayer.property("myProperty2", defaultValue = 1f)
 
 // Retrieves Int property or null if the property does not exist.
 val myOtherProp: Int? = map.propertyOrNull("myOtherProperty")
@@ -200,8 +200,8 @@ if(map.contains("enemyLayer")) {
 }
 
 // or with the "in" syntax
-if("collision" in map) {
-    val collisionLayer = map.layer("collision")    
+if ("collision" in map) {
+    val collisionLayer = map.layer("collision")
 }
 
 // the next line will throw a MissingLayerException if the layer does not exist
@@ -232,7 +232,8 @@ val map: TiledMap = getTiledMap()
 
 // Iterate over all object layers and parse them.
 // Note that parseObjectLayer is only called for layers of exact type MapLayer.
-// TiledMapTileLayer which is a subclass of MapLayer will not match.
+// For example, TiledMapTileLayer - which is a subclass of MapLayer - does not 
+// have this exact type and will not match.
 map.forEachLayer<MapLayer> { layer ->
     parseObjectLayer(layer)
 }
@@ -242,19 +243,18 @@ Check if a `MapLayers` or `MapObjects` collection is empty:
 
 ```kotlin
 import com.badlogic.gdx.maps.MapObjects
-import com.badlogic.gdx.maps.MapLayers
 import com.badlogic.gdx.maps.tiled.TiledMap
 import ktx.tiled.*
 
-val map: TiledMap = getTiledMap()
+val map: TiledMap = TiledMap()
 
-if(tiledMap.layers.isNotEmpty()) {
-    tiledMap.layers.forEach { layer ->
-        if(layer.objects.isEmpty()) {
+if (map.layers.isNotEmpty()) {
+    map.layers.forEach { layer ->
+        if (layer.objects.isEmpty()) {
             // nothing to do if there are no objects
             return@forEach
         }
-        
+
         parseObjects(layer.objects)
     }
 }
--- a/tiled/README.md
+++ b/tiled/README.md
@@ -73,8 +73,15 @@ the following extensions were added:
 - `contains(layerName: String)`: works as the `in` operator.
 - `layer(layerName: String)`: returns the layer or throws a `MissingLayerException` in case the layer does not exist.
 
-Inlined `forEachMapObject` extension methods allows to iterate over all `MapObject` instances present on the chosen
+Inlined `forEachMapObject` extension method allows to iterate over all `MapObject` instances present on the chosen
 map layer.
+
+Inlined `forEachLayer` extension method allows to iterate over all `MapLayer` instances of a specific type to execute
+a certain function on them.
+
+### `MapLayers` and `MapObjects`
+
+`isEmpty` and `isNotEmpty` extension method to check if the specific collection is empty or not.
 
 ### Usage examples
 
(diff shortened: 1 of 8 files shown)


tiled/README.md
```diff
@@ -215,6 +222,44 @@ map.forEachMapObject("collision") { mapObj ->
...
+
+```kotlin
+import com.badlogic.gdx.maps.MapObjects
+import com.badlogic.gdx.maps.MapLayers
+import com.badlogic.gdx.maps.tiled.TiledMap
+import ktx.tiled.*
+
+val map: TiledMap = getTiledMap()
+
+if(tiledMap.layers.isNotEmpty()) {
```
czyzby, line 251: Format the examples with the default KTX formatter.

tiled/README.md
```diff
@@ -215,6 +222,44 @@ map.forEachMapObject("collision") { mapObj ->
...
+
+```kotlin
+import com.badlogic.gdx.maps.tiled.TiledMap
+import ktx.tiled.*
+
+val map: TiledMap = getTiledMap()
+
+// Iterate over all object layers and parse them.
+// Note that parseObjectLayer is only called for layers of exact type MapLayer.
+// TiledMapTileLayer which is a subclass of MapLayer will not match.
```
czyzby, line 235: I'd paraphrase this to make it a bit clearer:

> For example, TiledMapTileLayer - which is a subclass of MapLayer - does not have this exact type and will not match.

tiled/src/main/kotlin/ktx/tiled/mapLayers.kt
```diff
@@ -38,3 +39,13 @@ inline fun <reified T> MapLayer.propertyOrNull(key: String): T? = properties[key
  * @return true if the property exists. Otherwise false.
  */
 fun MapLayer.containsProperty(key: String) = properties.containsKey(key)
+
+/**
+ * Returns **true** if and only if the [MapLayers] collection is empty.
+ */
+fun MapLayers.isEmpty() = this.count <= 0
```
czyzby, line 46: Is it possible for the `count` to be negative? Is that some special state that should be handled separately or just defensive programming?

tiled/src/main/kotlin/ktx/tiled/tiledMaps.kt
```diff
@@ -153,3 +154,18 @@ inline fun TiledMap.forEachMapObject(layerName: String, action: (MapObject) -> U
...
 }
+
+/**
+ * Extension method to run a [lambda] on a specific type of [layers][MapLayer] of the [TiledMap]. The lambda
+ * takes the matching [MapLayer] as a parameter.
+ *
+ * The class matching is an exact match meaning that a given subclass like [TiledMapTileLayer] will not match
+ * for the argument [MapLayer].
+ */
+inline fun <reified T : MapLayer> TiledMap.forEachLayer(lambda: (T) -> Unit) {
```
czyzby, line 165: Rename lambda to something along the lines of `operation` or `action`. `forEach` in stdlib uses `action`.

tiled/src/main/kotlin/ktx/tiled/tiledMaps.kt
```diff
@@ -153,3 +154,18 @@ inline fun TiledMap.forEachMapObject(layerName: String, action: (MapObject) -> U
...
+  this.layers.forEach {
+    if (it::class == T::class) {
```
czyzby, line 167: Is there a specific reason why you're using an exact class match? Is it like in the example with `MapLayer`, where it is used both for extension and as a concrete class for specific layers?

tiled/src/main/kotlin/ktx/tiled/mapObjects.kt
```diff
@@ -120,3 +121,13 @@ val MapObject.shape: Shape2D
     is RectangleMapObject -> rectangle
     else -> throw MissingShapeException("MapObject of type ${this::class.java} does not have a shape.")
   }
+
+/**
+ * Returns **true** if and only if the [MapObjects] collection is empty.
+ */
+fun MapObjects.isEmpty() = this.count <= 0
```
czyzby, line 128: If it's possible for the tiled map to return a null instead of `MapObjects`, this will be a bit more useful:
```kotlin
@OptIn(ExperimentalContracts::class)
fun MapObjects?.isEmpty(): Boolean {
  contract {
    returns(false) implies (this@isEmpty != null)
  }
  return this == null || this.count <= 0
}
```
This function, along with `isNotEmpty`, would work both as a null check and a count check. If it's never null, feel free to resolve this comment. Otherwise, the same applies to `isNotEmpty` and map layers.

tiled/src/test/kotlin/ktx/tiled/MapLayerTest.kt
```diff
@@ -44,4 +45,34 @@ class MapLayerTest {
...
+    assertTrue(actual.isEmpty())
+  }
+
+  @Test
+  fun `should return false when MapLayers is not empty`() {
+    val actual = MapLayers()
+    actual.add(MapLayer())
+
+    assertFalse(actual.isEmpty())
+  }
```
czyzby, line 62: This might be stretching it a bit and testing the internal implementation, but if there's a possibility of removing a layer, I'd also add a test where you add and remove the layer to see if the count is correctly updated and the extension methods work properly.

I feel like some additional tests that are borderline for LibGDX itself are a good thing, considering they don't have much of a test suite as far as I know.

czyzby (commented):
Nice work overall, might require some slight adjustments.

czyzby:
Just a heads up - I've updated the Kotlin formatter along with other dependencies, so you might want to pull the changes and reformat the code. If you're using the official formatter for the project in your IDE, you can also run `linterIdeSetup` Gradle task to update the config.

tiled/src/main/kotlin/ktx/tiled/mapLayers.kt
Quillraven: It simply returns the underlying size of the array. I now just saw that there is also a `size()` function which does exactly the same and has the same comment as `getCount()`.

Shall I convert it to == 0? I always do <= 0 because it is the exact opposite of > 0. But no, the value can never be negative in my opinion.

tiled/src/main/kotlin/ktx/tiled/tiledMaps.kt
Quillraven: Tiled has four specific layer types currently:
- Tiled
- Object
- Image
- Group

For whatever reason LibGDX only implemented three of them: `MapGroupLayer`, `TiledMapImageLayer` and `TiledMapTileLayer`. And they have a generic `MapLayer` class.

The way I needed this extension was that I wanted to iterate over Object layers, get their objects and create entities out of it. Since the object layer is just a generic `MapLayer` I would get all layers again if I use an instance check.

Also, when you use `forEachLayer` you most likely want to do that for one specific layer type.

image

tiled/src/main/kotlin/ktx/tiled/mapObjects.kt
Quillraven: A `Map` in LibGDX always comes with an initialized `MapLayers` and `MapProperties` object. There is no setter from what I can tell. Therefore, null should not be possible for `MapLayers` unless you hack yourself in with reflection.

The same is true for `MapLayer` and its `MapObjects` property.

Should we consider the reflection use case? I personally wouldn't do it because when you hack around like that you are on your own ;)

```Java
public class Map implements Disposable {
	private MapLayers layers = new MapLayers();
	private MapProperties properties = new MapProperties();
```

```Java
public class MapLayer {
	private String name = "";
	// ...
	private MapObjects objects = new MapObjects();
	private MapProperties properties = new MapProperties();
```

tiled/src/test/kotlin/ktx/tiled/MapLayerTest.kt
Quillraven: I added that test case. I don't know if people ever add/remove layers manually. The way I use it is that the entire map from Tiled gets loaded with the TmxMapLoader of LibGDX. That loader prepares a TiledMap and all its content and I never manipulated the map data during runtime.

However, it is possible and maybe people out there do it, so it makes sense to have a test for that.

Quillraven:
Pulled from the upstream and hopefully updated the formatting correctly :)

tiled/src/main/kotlin/ktx/tiled/mapLayers.kt
czyzby: @Quillraven I was just curious, I wouldn't be surprised if `-1` had some special meaning.

tiled/src/main/kotlin/ktx/tiled/mapObjects.kt
czyzby: > Should we consider the reflection use case? I personally wouldn't do it because when you hack around like that you are on your own ;)

No, it's OK. No need to be paranoid. :)

tiled/src/test/kotlin/ktx/tiled/MapLayerTest.kt
czyzby: Thanks.

czyzby (changes requested):
Looks good now, thanks.

czyzby (approved):
Misclicked the previous review. It was supposed to be an approval. :)

Quillraven:
cool :)

czyzby:
These additions should be available as a snapshot shortly.

@Quillraven By the way, I might be a little harsh during the reviews sometimes with the null checks and whatnot, but my goal is not to be paranoid about Java APIs. A few additional tests for edge cases are OK, but I don't want to go overboard with checks that would only make sense e.g. if reflection is involved. Some of my comments are just what-ifs to avoid shipping code with bugs under some weird conditions. I hope you're OK with that. :)

Quillraven:
Totally fine for me and I actually think the opposite. I take it as a very valuable input and it let's me think again about some solutions.

I learned already a lot from you  :)

Issue #398, opened 2021-12-04
Selection listener for nodes in Trees


manabreak:
Apologies if this feature is already in place; I did browse through the code but couldn't find this feature yet.

I was wondering if it would be possible to add a selection listener for Tree objects. I was thinking of this kind of format:

```
scene2d.visTree {
    visLabel("Foo")
    visLabel("Bar") { node ->
        visLabel("Baz")
    }
    onSelection { node ->
        // Invoked when a node in the tree was clicked
    }
}
```

czyzby:
I don't recall such an extension. If libGDX `Tree` allows to register listeners, it's certainly possible to add this utility.

If you'd like to contribute it, don't hesitate to make a PR. Otherwise I'll look into it in about 2 weeks.

manabreak:
Good to hear! I might look into it this weekend, if I have time.

czyzby:
Thanks for the contribution, @manabreak. The extension method was renamed to `onSelectionChange` and should be available via a snapshot release soon.

PR #399, opened 2021-12-04
Tree.onItemClick support, #398

manabreak:
- Added support for onItemClick listeners for Trees

### DIFF
--- a/actors/src/main/kotlin/ktx/actors/events.kt
+++ b/actors/src/main/kotlin/ktx/actors/events.kt
@@ -7,6 +7,7 @@ import com.badlogic.gdx.scenes.scene2d.InputEvent.Type.keyDown
 import com.badlogic.gdx.scenes.scene2d.InputEvent.Type.keyTyped
 import com.badlogic.gdx.scenes.scene2d.InputEvent.Type.keyUp
 import com.badlogic.gdx.scenes.scene2d.InputListener
+import com.badlogic.gdx.scenes.scene2d.ui.Tree
 import com.badlogic.gdx.scenes.scene2d.utils.ChangeListener
 import com.badlogic.gdx.scenes.scene2d.utils.ChangeListener.ChangeEvent
 import com.badlogic.gdx.scenes.scene2d.utils.ClickListener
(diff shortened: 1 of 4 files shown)


actors/src/main/kotlin/ktx/actors/events.kt
```diff
@@ -88,6 +89,20 @@ inline fun <T : Actor> T.onClickEvent(
   return clickListener
 }
 
+/**
+ * Attaches a [ClickListener] to this tree.
+ * @param lsitener invoked each time a node in this tree is clicked. Consumes the triggered [InputEvent] and the
+ * [Tree] the listener was originally attached to as `this`. The received Node is the node that was clicked.
+ * @return [ClickListener] instance.
+ */
+inline fun <T : Tree<*, *>> T.onItemClick(crossinline listener: ((Tree.Node<*, *, *>) -> Unit)): ClickListener {
```
czyzby, line 98: You might be able to add more generic types so that the listener receives a specific type of node, but I'm not sure how useful that would be in this case. Depends on whether the tree is constructed via our DSL or manually with a custom node class.

czyzby: You can try to introduce more generic type, so that there's no need to use wildcard types for node, but if it proves cumbersome to use, then feel free to scrap it.

actors/src/main/kotlin/ktx/actors/events.kt
```diff
@@ -88,6 +89,20 @@ inline fun <T : Actor> T.onClickEvent(
...
+  val clickListener = object : ClickListener() {
```
czyzby, line 99: Tree's internal `ClickListener` seems to fire a change listener when the selection changes. It might make sense to use a change listener instead, but make sure to rename the entire function (e.g. `onSelectionChange`).

actors/src/main/kotlin/ktx/actors/events.kt
```diff
@@ -88,6 +89,20 @@ inline fun <T : Actor> T.onClickEvent(
...
+    override fun clicked(event: InputEvent?, x: Float, y: Float) = listener(selectedNode)
```
czyzby, line 100: `getSelectedNode` returns only the first selected node, and can return null. It might be more appropriate to return the entire `Selection` object or a list of selected nodes returned by it.

actors/src/test/kotlin/ktx/actors/EventsTest.kt
```diff
@@ -79,6 +81,19 @@ class EventsTest {
     assertTrue(listener in actor.listeners)
   }
 
+  @Test
+  fun `should attach ClickListener to a Tree consuming InputEvent with clicked Node`() {
```
czyzby, line 85: Please add a test that verifies the selection content (i.e. simulate actual `Tree` usage). Verifying if the listener is in an actor is good first step, but since this logic is specific to a certain widget, we can do more than that to make sure it works correctly.

czyzby (changes requested):
Thanks for the contribution. Overall, I think we should use `ChangeListener` that accepts either a list(/`GdxArray`) of selected nodes or a `Selection` object.

czyzby:
@manabreak Hey, are you still working on this? If you don't have the time to apply the changes, I can merge this as is and refactor the code myself.

czyzby:
I will merge this as is and apply the fixes locally. Thanks for your contribution, @manabreak.

czyzby (approved):

PR #403, opened 2022-01-18
Update libGDX wiki links in README

crykn:
Fixes the libGDX wiki links which were broken in the course of the recent [wiki migration](https://libgdx.com/news/2022/01/devlog-8-wiki-migration). Similarily, the KTX wiki also has some links that need fixing.


czyzby:
Thanks @crykn. Would you also be able to fix the links in nested `README` files of each KTX module? They are usually at the end of the files.

crykn:
Done. Please excuse the commit mess, but I have no idea how to force push via GitHub's online interface (if that is even possible).

czyzby:
It's OK. Thanks a lot for taking the time to do this!

Issue #422, opened 2022-07-26
Using property delegates for Ashley mappers


czyzby:
The goal of this task is to add property delegates to `ktx-ashley` module to support accessing components via entity properties. For example

```kotlin
class MapperDelegate<T : Component>(val mapper: ComponentMapper<T>) {
    operator fun getValue(thisRef: Entity, property: KProperty<*>): T {
        return mapper[thisRef]
    }

    operator fun setValue(thisRef: Entity, property: KProperty<*>, value: T) {
        thisRef.add(value)
    }
}

// Declaration
data class Location(var x: Int = 0, var y: Int = 0) : Component
val LocationMapper = mapperFor<Location>()
var Entity.location: Location by MapperDelegate(LocationMapper)
// Or:
var Entity.location: Location by mapperPropertyFor<Location>()

// Usage example:
val engine = Engine()
val e = Entity()
engine.addEntity(e)
e.add(Location(3,2))

println(e.location.x) // prints 3
```

<sup>Originally posted by **@pylvain** July 26, 2022, see #421</sup>

czyzby:
@pylvain I've added some documentation and tests to the component property delegates that you've implemented. Thanks again for the suggestion and the pull request! These utilities should be available shortly via the snapshot version, and I might consider releasing a stable version soon with support for a newer Kotlin release.

PR #423, opened 2022-07-27
Added delegated properties for the Entity class

pylvain:
#422

### DIFF
--- a/ashley/src/main/kotlin/ktx/ashley/properties.kt
+++ b/ashley/src/main/kotlin/ktx/ashley/properties.kt
@@ -1,0 +1,45 @@
+import com.badlogic.ashley.core.Component
+import com.badlogic.ashley.core.ComponentMapper
+import com.badlogic.ashley.core.Entity
+import ktx.ashley.mapperFor
+import kotlin.reflect.KProperty
+
+/**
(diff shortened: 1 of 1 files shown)


czyzby:
Thanks for the contribution, @pylvain. That's a quite elegant extension for the `ktx-ashley` module!

I can see that the Kotlin file is not formatted properly - can you run the `format` Gradle task (e.g. `./gradlew ashley:format` in command line), so that the module is reformatted?

We also try to provide unit tests and update the [change log](https://github.com/libktx/ktx/blob/master/CHANGELOG.md) for every change within the library. `ktx-ashley` is a bit unique in this regard, since it uses Spek for tests instead of JUnit. If you're unfamiliar with the Spek framework or do not have time to write the unit tests, I can handle this after merging.

Just let me know if you'd like to finish the contribution or whether you'd like me to take over. Thanks.

pylvain:
Well, It was a pleasure having this conversation with you. Maybe I'll get to contribute to this library again since I'm working with it. 
I'm not familiar with unit testing, contributing was already new for me, I'll pass on this one, you can take over ! Thank you.

czyzby:
Thanks for running the formatter. I'll link the related commits with #422, so you can check out how the tests are implemented. Thanks again for the contribution!

PR #426, opened 2022-07-30
Adding TagDelegate and tagFor

pylvain:
#427

### DIFF
--- a/ashley/src/main/kotlin/ktx/ashley/properties.kt
+++ b/ashley/src/main/kotlin/ktx/ashley/properties.kt
@@ -121,12 +121,13 @@ inline fun <reified T : Component> tagFor(noinline defaultValueProvider: () -> T
 
 inline fun <reified T : Component> tagFor(noinline defaultValueProvider: (() -> T)? = null): TagDelegate<T> {
   if (defaultValueProvider != null) {
-    return TagDelegate(mapperFor(), T::class.java,defaultValueProvider)
+    return TagDelegate(mapperFor(), T::class.java, defaultValueProvider)
   }
   val defaultConstructorIndex = T::class.java.constructors.indexOfFirst { it.parameters.isEmpty() }
   if (defaultConstructorIndex >= 0) {
     return TagDelegate(mapperFor(), T::class.java) {
-      T::class.java.constructors[defaultConstructorIndex].newInstance() as T }
+      T::class.java.constructors[defaultConstructorIndex].newInstance() as T
+    }
   }
   return TagDelegate(mapperFor(), T::class.java, null)
 }
--- a/ashley/src/main/kotlin/ktx/ashley/properties.kt
+++ b/ashley/src/main/kotlin/ktx/ashley/properties.kt
@@ -78,21 +78,22 @@ inline fun <reified T : Component> optionalPropertyFor(
 /**
  * Property delegate for an [Entity] wrapping around a [ComponentMapper].
  * Allows checking the presence of a component on an Entity.
- * Allows assigning the defaultValue argument and removing a component from an Entity with a Boolean value
+ * Allows assigning the result of the invocation of defaultValueProvider to the entity
+ * and removing a component from an Entity with a Boolean value
+ * Attempting to set a value when defaultValueProvider is null will throw a [NullPointerException]
  */
 
 class TagDelegate<T : Component>(
   val mapper: ComponentMapper<T>,
   private val componentClass: Class<T>,
-  private val defaultValue: T
+  private val defaultValueProvider: (() -> T)?
 ) {
-
   operator fun getValue(thisRef: Entity, property: KProperty<*>): Boolean =
     mapper.has(thisRef)
 
   operator fun setValue(thisRef: Entity, property: KProperty<*>, value: Boolean) {
     if (value) {
-      thisRef.add(defaultValue)
+      thisRef.add(defaultValueProvider!!.invoke())
     } else {
       thisRef.remove(componentClass)
     }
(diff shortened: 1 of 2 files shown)


ashley/src/main/kotlin/ktx/ashley/properties.kt
```diff
@@ -74,3 +74,38 @@ inline fun <reified T : Component> optionalPropertyFor(
...
+ */
+
+class TagDelegate <T : Component> (val mapper: ComponentMapper<T>, private val componentClass: Class<T>) {
+
+  operator fun getValue(thisRef: Entity, property: KProperty<*>): Boolean =
+    mapper.has(thisRef)
+
+  operator fun setValue(thisRef: Entity, property: KProperty<*>, value: Boolean) {
+    if (value) {
+      thisRef.add(componentClass.getConstructor().newInstance())
```
czyzby, line 91: Using reflection is tricky. This will result in runtime exceptions if there is no no-arg constructor (which isn't a good use case for the `TagDelegate`, but still). In general, it would be better to provide either a singleton instance for immutable components or a component provider lambda if a singleton is not acceptable.

czyzby (commented):
I can merge this as is, but in general I'd like to avoid reflection usage - especially since those kinds of components are often immutable and a single instance is sufficient. Let me know if you'd like to refactor this or whether you'd prefer I take over.

ashley/src/main/kotlin/ktx/ashley/properties.kt
pylvain: Ashley.Ktx already have functionalities that require to have a no-arg constructor for Components and throw exception if not  ( The ``` engine.add { entity { with<Component> } } }``` does ), so I don't think it would be that bad.
I'd like to keep a ```tagFor``` function that works just with the type supplied, it's just so elegant. 
Maybe we can make the function tagFor throw an Exception if there is no no-arg constructor and no value supplied? 
The delegate tagFor could have a parameter that is the default value of the Component.
The tagFor function could have an optional argument that is the default value of the Component, and provide the value of the no-arg constructor by default, so the reflection operation only happens once at the creation of the property.

czyzby: Reflection should be avoided whenever possible. It's the most error-prone API on platforms such as iOS or experimental web, since they basically have to emulate JVM reflection. The performance might also vary from platform to platform - it's [pretty OK](https://stackoverflow.com/questions/435553/java-reflection-performance) on desktop (twice as slow, still kinda fast), but it might be worse on mobiles.

It's true that `ktx-ashley` provides functions such as `with` that rely on reflection but it also has a way to avoid them entirely with utilities such as the `+=` operator. For example, this syntax should be preferred for performance-critical applications:

```kotlin
  val entity = Entity()
  entity += Transform(x=4f, y=2f)

  // With engine:
  val engine = Engine()
  engine.entity { 
    entity += Transform(x=4f, y=2f)
  }
```

It's OK to provide a reflection-based solution to speed things up, but there should be an option to avoid it entirely.

pylvain: I think it's nice for prototyping, but we can mention in the doc that it is faster to provide a default Component value. In the last commit, the reflection operation happens only once per property declared, so It's not that great of deal.

pylvain:
After further testing, I think I want to force the user to specify a default value everytime. I did not think that with type inference, it would just look like:
```kotlin
var Entity.isVisible = tagFor(Visible())
```
which is perfectly fine IMO and avoids reflection. 
I will edit the pull request in consequence.

I was thinking of adding another delagated property that would be used just to check the existence of a component. Maybe i'll do another pr.

czyzby:
> I was thinking of adding another delagated property that would be used just to check the existence of a component. Maybe i'll do another pr.

I don't think that's necessary, since you can use a read-only `TagDelegate` to achieve just that.

czyzby:
> After further testing, I think I want to force the user to specify a default value everytime.

Ideally, I think it makes sense to provide both options, depending on the use case - the user might need either a singleton that is always reused or a provider that creates a new instance (through lambda or with reflection) depending on whether the component has any extra fields. You can apply your changes as you see fit and I'll likely merge this as is, but I might just add a few extensions to support more use cases.

pylvain:
> > I was thinking of adding another delagated property that would be used just to check the existence of a component. Maybe i'll do another pr.
> 
> I don't think that's necessary, since you can use a read-only `TagDelegate` to achieve just that.

You can use `TagDelegate` to achieve that only if the Component  has default value or no-arg constructor.
 
```kotlin
data class Location(x:Int,y:Int)
val Entity.hasLocation by tagFor<Location>() // throws an Exception since Location has non default value
```

I don't see a way around that using the tagFor() function, since the function cannot know if it will be assigned to a val or a var.

czyzby:
> I don't see a way around that using the tagFor() function, since the function cannot know if it will be assigned to a val or a var.

I have a couple ideas, lazy initiation should do the job. `TagDelegate` can store a `() -> T` reference that either returns a singleton or provides a new instance depending on what parameters were passed to `tagFor`.

pylvain:
Updated Pr to make the functions `tagFor` and `TagDelegate` accept a provider function instead of a default value.
Now tagFor works for readonly properties with no default constructor or value provided.
It's some boilerplate code, but maybe i'll make a delagate that accept a Singleton instead of a provider ?

czyzby:
> It's some boilerplate code, but maybe i'll make a delagate that accept a Singleton instead of a provider ?

You can overload `tagFor` with an option that accepts a singleton and creates a provider for that. You can also convert `TagDelegate` to an interface and provide 2 separate implementations (singleton vs provider) - that's probably what I'd do to make it slightly simpler.

pylvain:
Updated tagFor with Singleton/Provider Attributes
I'm 100% happy with the behavious of this one, except one thing I cannot manage to solve by myself :
```kotlin
var Entity.isLocated by tagFor { Location(0,0) }  // Must specify <Location> after tagFor, overwise overload resolution ambiguity
```
That's so confusing for the user it must be changed.
Obviously we can solve the problem with two functions singletonTagFor and providerTagFor but It would be better to have only one function tagFor.
This goes beyond my competence however. Maybe you can figure it out ?
I can try to write the tests after we're done with this pr if it's ok with you.

czyzby:
> This goes beyond my competence however. Maybe you can figure it out ?

I'll have a busy next few days, but I'll try to suggest something by the end of the week. I can also merge the PR then and update documentation (and so on), since it already looks pretty good.

>  I can try to write the tests after we're done with this pr if it's ok with you.

Let's keep this within a single PR. You can try to write the tests in the meanwhile, and I'll finish or refactor them if necessary. Just make sure that the pipeline passes. [Check out this commit](https://github.com/libktx/ktx/commit/3a02a28686245361d5c6b945c6987ecb6d4b65a6) to see the changes that I've done to `README`, `CHANGELOG` and the tests after your previous PR. Thanks.

pylvain:
Wrote some tests. I have little experience with that. Hope it will save you some time.
I was thinking of creating a ```propertyWithDefaultFor``` in the same spirit of ```tagProperty```, where a default value would be returned if the component doesn't exist supplied by a default value/singleton/provider/call to the no-arg constructor.
I'll do that when the problem with ```tagFor``` is solved.
I also think the functions ```tagFor``` and ```propertyFor``` should be the near the top for the readme file. IMO it's now the best way to interact with components, I think they deserve some visibility.

czyzby:
> I was thinking of creating a ```propertyWithDefaultFor``` in the same spirit of ```tagProperty```, where a default value would be returned if the component doesn't exist supplied by a default value/singleton/provider/call to the no-arg constructor.

I think that's a bit error-prone, since you can accidentally create a component whilst making another check (such as `entity.component != null`). A default value might be nice for some use cases, but I'm a bit afraid it might be used over the optional property to get around null checks. The potential risks would have to be documented. You can propose another PR for this functionality if you want, but please checkout and select the `develop` when setting up the branches.

>  I also think the functions tagFor and propertyFor should be the near the top for the readme file. IMO it's now the best way to interact with components, I think they deserve some visibility.

Noted, I'll move them above custom mappers examples.

czyzby:
Thanks for the PR. I'll merge it as is and adjust the tests to cover all use cases.

Issue #479, opened 2024-03-21
Question about new additions


Quillraven:
Hi,

I am currently doing another project with LibGDX/LibKTX and I missed a few extensions that I implemented now myself but I could also make them a contribution to LibKTX if needed.
The things I am missing:
- destructor operators for certain "math" shapes like Rectangle, Ellipse, ...
- a `use` extension for TiledMap renderer (similar to what we have for batch)

Is that something useful? Or do you think that should not be part of LibKTX?

Rectangle example:

```kotlin
operator fun Rectangle.component1(): Float = this.x
operator fun Rectangle.component2(): Float = this.y
operator fun Rectangle.component3(): Float = this.width
operator fun Rectangle.component4(): Float = this.height
```

czyzby:
Sure, these could be added to `ktx-math` and `ktx-tiled` respectively. Would you like to submit a PR?

Quillraven:
Yes, I will try to create one over the weekend :)

PR #480, opened 2024-03-24
#479: add math Shape2D destructure extensions and MapRenderer use extension

Quillraven:
PR for #479

Unfortunately, the `beginRender()` and `endRender()` methods of the `OrthogonalTiledMapRenderer` are protected, which means we cannot call them directly.
I never needed a custom renderer or had to override those methods. Also, they seem to never change in recent LibGDX versions. That's why I think "copy&pasting" their content is fine in this case. However, I still added a remark in the ReadMe.md that this can be a potential issue.

If you have a problem with it then feel free to remove those extensions :)

### DIFF
--- a/math/README.md
+++ b/math/README.md
@@ -36,6 +36,12 @@ used to quickly determine which velocity or force is greater. Note that length *
 much faster to calculate and yields the same results in most cases.
 - `dot` infix function allows to calculate the dot product of 2 vectors.
 - `x` infix function allows to calculate the cross product of 2 vectors.
+- `Rectangle` and `Ellipse` instances can be destructed to four float variables in one step with
+  `val (x, y, w, h) = rectangle/ellipse` syntax thanks to `component1()`, `component2()`, `component3()` and `component4()` operator methods.
+- `Circle` instances can be destructed to three float variables in one step with
+  `val (x, y, radius) = circle` syntax thanks to `component1()`, `component2()` and `component3()` operator methods.
+- `Polygon` and `Polyline` instances can be destructed to two float variables in one step with
+  `val (x, y) = polygon/polyline` syntax thanks to `component1()` and `component2()` operator methods.
 
 Note that since `Shape2D` has `contains(Vector2)` method, `in` operator can be used for any `Shape2D` implementation
 (like `Rectangle`, `Ellipse` or `Circle`). For example, given `vec: Vector2` and `rect: Rectangle` variables, you can
--- a/math/README.md
+++ b/math/README.md
@@ -36,12 +36,6 @@ used to quickly determine which velocity or force is greater. Note that length *
 much faster to calculate and yields the same results in most cases.
 - `dot` infix function allows to calculate the dot product of 2 vectors.
 - `x` infix function allows to calculate the cross product of 2 vectors.
-- `Rectangle` and `Ellipse` instances can be destructed to four float variables in one step with
-  `val (x, y, w, h) = rectangle/ellipse` syntax thanks to `component1()`, `component2()`, `component3()` and `component4()` operator methods.
-- `Circle` instances can be destructed to three float variables in one step with
-  `val (x, y, radius) = circle` syntax thanks to `component1()`, `component2()` and `component3()` operator methods.
-- `Polygon` and `Polyline` instances can be destructed to two float variables in one step with
-  `val (x, y) = polygon/polyline` syntax thanks to `component1()` and `component2()` operator methods.
 
 Note that since `Shape2D` has `contains(Vector2)` method, `in` operator can be used for any `Shape2D` implementation
 (like `Rectangle`, `Ellipse` or `Circle`). For example, given `vec: Vector2` and `rect: Rectangle` variables, you can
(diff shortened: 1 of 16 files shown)


math/README.md
```diff
@@ -36,6 +36,12 @@ used to quickly determine which velocity or force is greater. Note that length *
 much faster to calculate and yields the same results in most cases.
 - `dot` infix function allows to calculate the dot product of 2 vectors.
 - `x` infix function allows to calculate the cross product of 2 vectors.
+- `Rectangle` and `Ellipse` instances can be destructed to four float variables in one step with
+  `val (x, y, w, h) = rectangle/ellipse` syntax thanks to `component1()`, `component2()`, `component3()` and `component4()` operator methods.
+- `Circle` instances can be destructed to three float variables in one step with
+  `val (x, y, radius) = circle` syntax thanks to `component1()`, `component2()` and `component3()` operator methods.
+- `Polygon` and `Polyline` instances can be destructed to two float variables in one step with
+  `val (x, y) = polygon/polyline` syntax thanks to `component1()` and `component2()` operator methods.
```
czyzby, line 44: Please move those to a separate section, this one is related to the extensions for `Vector2`.

tiled/src/main/kotlin/ktx/tiled/batchTiledMapRenderer.kt
```diff
@@ -0,0 +1,74 @@
...
+ * Automatically calls [BatchTiledMapRenderer.beginRender] and [BatchTiledMapRenderer.endRender].
+ * @param camera A camera to set on the renderer before [BatchTiledMapRenderer.beginRender].
+ * @param block inlined. Executed after [BatchTiledMapRenderer.beginRender] and before [BatchTiledMapRenderer.endRender].
+ */
+inline fun <T : BatchTiledMapRenderer> T.use(
+  camera: OrthographicCamera,
+  block: (T) -> Unit
+) {
+  this.setView(camera)
+  AnimatedTiledMapTile.updateAnimationBaseTime()
```
czyzby, line 19: I'm wondering whether this update should be locked under a boolean flag such as `updateAnimation`. Could default to `true` if this is always expected to be called before using the renderer.

czyzby:
> Unfortunately, the `beginRender()` and `endRender()` methods of the `OrthogonalTiledMapRenderer` are protected, which means we cannot call them directly.

There's a cheesy way to get around this. Create a separate Kotlin file, set the package to match `OrthogonalTiledMapRenderer` (with a warning suppression that is doesn't match the folder structure :wink:), make utility functions or extension methods that call the `beginRender()` and `endRender()`, and finally use those in the `use` extension. This would call any overridden methods.

tiled/src/main/kotlin/ktx/tiled/batchTiledMapRenderer.kt
Quillraven: It is called per default in LibGDX and from my understanding it is linked to the animated tiles that you can create in Tiled map editor. I personally find it very weird that this is included inside a `beginRender` method but I guess they expected that people will just call `render` most of the time which renders the entire map and includes one update of the animation timer, which does make sense.

However, imo, most games will split between background and foreground layer rendering which means you need to take care of `batch.begin`, `batch.end` yourself and at least I fell into the trap once that I forgot to call this static AnimatedTileMapTile update method :D

Anyway, to summarize, imo it does not hurt to have it as a parameter with default value true. I will include it :)

Quillraven: edit: nevermind; if we make the cheesy solution of yours then I anyway cannot control anymore if it gets called or not because it is part of `beginRender` of the map renderer implementation ;)

math/README.md
Quillraven: I moved it to the end of the file

Quillraven:
> > Unfortunately, the `beginRender()` and `endRender()` methods of the `OrthogonalTiledMapRenderer` are protected, which means we cannot call them directly.
> 
> There's a cheesy way to get around this. Create a separate Kotlin file, set the package to match `OrthogonalTiledMapRenderer` (with a warning suppression that is doesn't match the folder structure 😉), make utility functions or extension methods that call the `beginRender()` and `endRender()`, and finally use those in the `use` extension. This would call any overridden methods.

I did that trick now. Did not know about it and feels weird that this is even possible :) But a nice solution imo for this scenario.

tiled/README.md
```diff
@@ -83,6 +83,17 @@ a certain function on them.
 
 `isEmpty` and `isNotEmpty` extension method to check if the specific collection is empty or not.
 
+### `BatchTiledMapRenderer`
+
+`use` extension method to call `beginRender()` and `endRender()` automatically before
+your render logic.
+
+**Careful**: Since `beginRender()` and `endRender()` are protected methods, they are
```
czyzby, line 91: Is this still relevant? I think this section can be removed.

tiled/src/test/kotlin/ktx/tiled/TiledMapRendererTest.kt
```diff
@@ -0,0 +1,75 @@
...
+  @Test
+  fun `should call beginRender and endRender without setView`() {
+    val renderer = mock<BatchTiledMapRenderer>()
+
+    renderer.use {
+      verify(renderer).beginInternal()
+      assertSame(renderer, it)
+      verify(renderer, never()).endInternal()
+    }
+    verify(renderer).endInternal()
```
czyzby, line 27: Not sure if you can verify calls on essentially static methods, you might have to change the tests package as well to verify the protected methods.

tiled/README.md
Quillraven: No, that can be removed. I forgot that, sorry

tiled/src/test/kotlin/ktx/tiled/TiledMapRendererTest.kt
Quillraven: Why is that necessary? Isn't it sufficient to verify if the wrapper functions got called? Because the actual functions cannot be verified since they are protected. At least to my knowledge it is not advisable to verify protected/private methods.

czyzby: I thought it might have been a limitation of Mockito, but apparently not. The tests pass, so I'm fine with verifying the extension methods.

czyzby:
@Quillraven I missed that you've set the `master` as the target branch instead of `develop`. Can you set another PR to the `develop` branch? Sorry!

Issue #485, opened 2024-06-22
scene2d image without drawable


Quillraven:
Hi,

right now the `image` scene2d extension requires a non-null `Drawable`:

```kotlin
@Scene2dDsl
@OptIn(ExperimentalContracts::class)
inline fun <S> KWidget<S>.image(
  drawable: Drawable,
  init: (@Scene2dDsl Image).(S) -> Unit = {},
): Image {
  contract { callsInPlace(init, InvocationKind.EXACTLY_ONCE) }
  return actor(Image(drawable), init)
}
```

However, it is valid to pass "null" as a drawable and this can be useful if the real drawable is set later on but you still want to add the image to the table to have the correct layout.

Would it be okay to make it `Drawable?` instead of `Drawable`? This might be also true for other extensions with drawables but I did not check them all.

Or is there a different solution to it?

czyzby:
If an image without a drawable can exist and might make sense for some use cases, it should be possible to create with the DSL. This might be an oversight, but it can be easily amended without breaking the API. `Drawable?` with a `null` default could be a way to solve this, I just have to check first if an image without a drawable isn't otherwise broken (e.g., thowing NPE on draw). Thanks for bringing this up.

Quillraven:
Imo this should be just fine and at least in my case it works. Here is also the official documentation of `setDrawable` which mentions it may be null:

![image](https://github.com/libktx/ktx/assets/93260/604c9e16-17d7-492d-badf-2e9a3b66a5d6)

Quillraven:
Will be fixed by the linked PR above

PR #487, opened 2024-07-29
(#485) make drawable and drawableName optional for scene2d image factory

Quillraven:
PR for #485.

I did not adjust the factory methods for Texture, TextureRegion and NinePatch because I think they will throw an exception, but more importantly, I think in those cases it does not make sense. Drawable and drawable name do make sense because in a Skin, from what I know, you can reference Drawables and in the case of an image they are optional.

### DIFF
--- a/scene2d/src/main/kotlin/ktx/scene2d/factory.kt
+++ b/scene2d/src/main/kotlin/ktx/scene2d/factory.kt
@@ -209,7 +209,7 @@ inline fun <S> KWidget<S>.horizontalGroup(
 }
 
 /**
- * @param drawableName optional name of a drawable stored in the chosen skin.
+ * @param drawableName name of a drawable stored in the chosen skin.
  * @param skin [Skin] instance that contains the widget style. Defaults to [Scene2DSkin.defaultSkin].
  * @param init will be invoked with the widget as "this". Consumes actor container (usually a [Cell] or [Node]) that
  * contains the widget. Might consume the actor itself if this group does not keep actors in dedicated containers.
@@ -219,16 +219,12 @@ inline fun <S> KWidget<S>.horizontalGroup(
 @Scene2dDsl
 @OptIn(ExperimentalContracts::class)
 inline fun <S> KWidget<S>.image(
-  drawableName: String? = null,
+  drawableName: String,
   skin: Skin = Scene2DSkin.defaultSkin,
   init: (@Scene2dDsl Image).(S) -> Unit = {},
 ): Image {
   contract { callsInPlace(init, InvocationKind.EXACTLY_ONCE) }
-  return if (drawableName == null) {
-    actor(Image(), init)
-  } else {
-    actor(Image(skin.getDrawable(drawableName)), init)
-  }
+  return actor(Image(skin.getDrawable(drawableName)), init)
 }
 
 /**
--- a/scene2d/src/main/kotlin/ktx/scene2d/factory.kt
+++ b/scene2d/src/main/kotlin/ktx/scene2d/factory.kt
@@ -209,7 +209,7 @@ inline fun <S> KWidget<S>.horizontalGroup(
 }
 
 /**
- * @param drawableName name of a drawable stored in the chosen skin.
+ * @param drawableName optional name of a drawable stored in the chosen skin.
  * @param skin [Skin] instance that contains the widget style. Defaults to [Scene2DSkin.defaultSkin].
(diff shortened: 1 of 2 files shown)


scene2d/src/main/kotlin/ktx/scene2d/factory.kt
```diff
@@ -219,12 +219,16 @@ inline fun <S> KWidget<S>.horizontalGroup(
 @Scene2dDsl
 @OptIn(ExperimentalContracts::class)
 inline fun <S> KWidget<S>.image(
-  drawableName: String,
+  drawableName: String? = null,
```
czyzby, line 222: Doesn't it leave 2 `image` methods that potentially have no parameters? If both have essentially the same effect, maybe let's leave the one with the drawable name as is, and make the `Drawable` argument nullable in the other one alone?

scene2d/src/test/kotlin/ktx/scene2d/factoryTest.kt
```diff
@@ -6,10 +6,7 @@ import com.badlogic.gdx.scenes.scene2d.ui.Cell
 import com.badlogic.gdx.scenes.scene2d.ui.Label
 import com.badlogic.gdx.scenes.scene2d.ui.Tree.Node
 import com.kotcrab.vis.ui.VisUI
-import org.junit.Assert
-import org.junit.Assert.assertEquals
-import org.junit.Assert.assertNotNull
-import org.junit.Assert.assertTrue
+import org.junit.Assert.*
```
czyzby, line 9: Please don't use wildcard imports, thanks.

scene2d/src/test/kotlin/ktx/scene2d/factoryTest.kt
```diff
@@ -142,7 +151,15 @@ class NoInitBlockActorFactoriesTest : ApplicationTest() {
   fun `should create Image with drawable`() = test(
     widget = { image(VisUI.getSkin().getDrawable("button")) },
     validate = {
-      Assert.assertSame(VisUI.getSkin().getDrawable("button"), it.drawable)
+      assertSame(VisUI.getSkin().getDrawable("button"), it.drawable)
+    },
+  )
+
+  @Test
+  fun `should create Image with null drawable`() = test(
```
czyzby, line 159: I'd appreciate it if you added another test for `image()` without a parameter just to see that it compiles and creates an image without a drawable.

czyzby (changes requested):

czyzby:
Thanks for the PR, I left some minor comments.

scene2d/src/main/kotlin/ktx/scene2d/factory.kt
Quillraven: Will check that. I had a similar concern about JVM that we end up with two methods of the same signature but I think that's not the case because on method takes 3 arguments and the other one 2. Not quite sure what kind of methods Kotlin compiler is creating with the default null argument but I think it should be fine, otherwise the IDE most likely would have shown me an alert 😅

Anyway, I will adapt it to your suggestion. That should work as well.

scene2d/src/test/kotlin/ktx/scene2d/factoryTest.kt
Quillraven: Will fix that too. This is done automatically for me I think when more than 5 imports of the same package are imported. Then IntelliJ makes an asterisk import for me.

Is this something that can be configured in .editorconfig file of the project?

czyzby: I think editor config is more language-agnostic than than, but I'd have to check.

Quillraven:
Issues should be resolved now. Regarding .editorconfig: I had a quick look and I could only find that there are IntelliJ specific properties that are supported. I guess the "star_import" properties are the one that would be needed. But since they have the prefix "ij_" it will most likely only work in IntelliJ.

image

The thing that I did to solve it was a setting directly in IntelliJ:
image

The default setting was "Use import with '*' when at least 5 names used / 3 names used

czyzby (approved):

czyzby:
Thanks, the changes should be available in the snapshot release shortly. I'll try to update the changelog and possibly push a new release over the weekend.

czyzby:
Just wanted to say that automatic snapshot upload failed, since I had to regenerate an access token into Maven central repository. Should be fixed now.

Quillraven:
Haha had the same issue with Flecks the last time. Welcome to the club 😉

Issue #499, opened 2025-05-13
Update to libGDX 1.13.5


czyzby:
See the [announcement](https://libgdx.com/news/2025/05/gdx-1-13-5).

czyzby:
Thank you @Quillraven for addressing this so quickly, the snapshot release with 1.13.5 support should be available shortly.

PR #500, opened 2025-05-16
Update LibGDX to 1.13.5, add TextMapObject to shape extension and add new forEachCell extension

Quillraven:


Quillraven:
seems like box2d extensions are broken because of the 1.13.5 update. Don't know what has changed there tbh

Quillraven:
Box2D test issues are resolved now. I also increased the fault tolerance of the profiling tests. On my machine I had 0.016f as a value which made the test fail. That's why I simply updated the tolerance. Shouldn't be a big deal.

----

Regarding Box2D:
There were some changes for 1.13.5: https://github.com/libgdx/libgdx/pull/7579
From what I understood you need platform specific natives for that to work which you have per default in a typical LibGDX app. But the tests in KTX were missing that dependency for the tests. Also, since we don't have an ApplicationListener in those tests, we need to manually call `GdxNativesLoader`.

Quillraven:
@czyzby: can you please have a look at the failing github action? I run gradle `test` on my machine and all tests were passing. So should be fine but the gradle wrapper validation is failing here?

czyzby:
Gradle wrapper validation fails sometimes, I reran the jobs. This should address #499, right? Thanks for submitting a PR for this @Quillraven!

After libGDX updates I tend to bump the version in a couple of additional places like the README (see https://github.com/libktx/ktx/commit/f45df779a180ed2dd153c6178b9210b8ecf44f94), but I can merge this as is and push an extra commit. I'll only ask you to update `version.txt` in the repository root to `1.13.5-SNAPSHOT` so that the latest snapshot release marked `1.13.1-SNAPSHOT` is not built against libGDX 1.13.5 as snapshots are pushed automatically. See [the contribution guide](https://github.com/libktx/ktx/blob/master/.github/CONTRIBUTING.md#updating-dependencies).

Quillraven:
#499 is resolved bis this PR as well, you are correct.

I updated some areas where I found 1.13.1-SNAPSHOT

czyzby:
Thanks!

Issue #281, opened 2020-04-26
Add TypingLabel extension module


czyzby:
- [ ] Create a new KTX module that extends `ktx-scene2d`.
- [ ] Include and document another major dependency: `com.rafaskoberg.gdx:typing-label`.
- [ ] Add and test factory method for `TypingLabel`.
- [ ] Add utility extension methods that register event listeners.
  - [ ] Add support for multiple `TypingListener` instances.
- [ ] Consider adding custom DSL for `TypingLabel` tokens.

---

[`TypingLabel`](https://github.com/rafaskb/typing-label) is an interesting third-party extension that defines additional Scene2D actor: a label that appears to be typed in real time. It supports [a wide range of special effects](https://github.com/rafaskb/typing-label/wiki/Tokens). It would be a nice addition to our Scene2D extensions.

![Label example](https://raw.githubusercontent.com/rafaskb/typing-label/master/media/effect-slide.gif)

michaeloa:
Is this one still on the menu? If so, I might consider looking into it, though I think I'd prefer to look in the direction of Textratypist (which I use myself) rather than TypingLabel.

czyzby:
@michaeloa I have no plans to work on it myself, but I'm happy to do code review and a release if you're like to tackle this one. Covering either library would be OK. Thanks!

michaeloa:
Cool. I've already done a bit of this work for my own use, since I use both, and wanted a simple way to use textratypist with ktx. I'll see if I can break out that code into it's own module and add some testing.

michaeloa:
Would appreciate some feedback on the draft PR before I finalize it.

Also, out of curiosity - any timeline on 1.14?

czyzby:
Thanks, I'll review it shortly.

> Also, out of curiosity - any timeline on 1.14?

I'm a bit behind on the updates, but I'll try to get them out before the PR is merged.

michaeloa:
I make games with lots of text, so a frequent use case of mine with textraLabel (and labels before that), is to do:

```
textraLabel("Some incredibly long text, etc") {
    alignment = Align.left,
    wrap = true
}
```

This gets annoying after a while, so I just end up creating a DSL function for this (currently `textraWrap`), since the above becomes quite unreadable after you're doing four or five of them in a single view. Would you be OK with me adding that type of factory function to the Textratypist extension? Or would you prefer to keep these types of convenience DSL outside of the extension modules?

PR #507, opened 2025-11-05
Add ktx-textratypist to libktx

michaeloa:
Contributing #281 : suport for [TextraTypist](https://github.com/tommyettinger/textratypist); adding animated and styled text labels such as TypingLabel and TextraLabel.


michaeloa:
First cut on implementing this. Having a little trouble with gradle format, so need to take a look at that (though hopefully the editorconfig has kept things relatively correct).

For now, I've just added support for TextraLabel and TypingLabel, the two main widgets. Not 100% sure about how prime-time the other widgets are, as I don't use them atm. But thinking it makes sense to start with a small(ish) commit, anyway.

Replaced VisUI with FWSkin in the tests, since I need support for Textra label styles.

Any obvious mistakes/issues that I need to address?

README.md
```diff
@@ -65,6 +65,7 @@ You can include the selected **KTX** modules based on the needs of your applicat
 |        [`ktx-scene2d`](scene2d)        | Type-safe Kotlin builders for [`Scene2D`](https://libgdx.com/wiki/graphics/2d/scene2d/scene2d) GUI.                               |
 |         [`ktx-script`](script)         | Kotlin scripting engine for desktop applications.                                                                                 |
 |          [`ktx-style`](style)          | Type-safe Kotlin builders for `Scene2D` widget styles extending `Skin` API.                                                       |
+|   [`ktx-textratypist`](textratypist)   | Type-safe Kotlin builders for [TextraTypist]() widgets.                                                                           |
```
czyzby: Please add a link to the library's repository.

buildSrc/src/main/kotlin/ktx/Versions.kt
```diff
@@ -7,7 +7,7 @@ const val artemisOdbVersion = "2.3.0"
 const val ashleyVersion = "1.7.4"
 const val gdxAiVersion = "1.8.2"
 const val visUiVersion = "1.5.5"
-
```
czyzby, line 10: I think that line might have been deliberate to separate test dependency versions. It's a tiny thing, but if you don't mind putting it back I'd appreciate it. :smile:

textratypist/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+TextraTypist requires a `Skin` with compatible styles for its widgets. This can be handled easily by using a `FWSkin`:
+
+```kotlin
+Scene2DSkin.defaultSkin = FWSkin(Gdx.files.internal("my-skin.json"))
+```
+
+## Available Builders
+
+| Widget | DSL Method                       | Description                                      |
```
czyzby: IntelliJ should have automatic formatting for Markdown tables, try reformatting the file to make it pretty even in text form.

textratypist/README.md
```diff
@@ -0,0 +1,66 @@
...
+}
+```
+
+### Using TextraTypist Markup Tags
+
+| Effect | Example Usage            |
+|-------|--------------------------|
+| Color | `[RED]Danger![]`         |
+| Wave Motion | `{WAVE}Ocean text[]`     |
+| Slow / Fast typing | `{SLOW}`, `{FAST}`, etc- |
```
czyzby: * Mention that `[]` clears the (color?) formatting.
  * Does `{}` do the same for animations?
* Leave a link to the complete list of markers.

textratypist/README.md
```diff
@@ -0,0 +1,66 @@
...
+```kotlin
+typingLabel("{WAVE}System reboot[] in [YELLOW]{SLOWER}3... 2... 1...[]")
+```
+
+### Synergy
+
+| Library | Purpose |
+|--------|---------|
+| `ktx-scene2d` | Layout + widget DSL |
+| `ktx-actors` | Stage helpers & utilities |
```
czyzby: READMEs typically also mention alternatives. I think the other typing label library should be mentioned here.

textratypist/gradle.properties
```diff
@@ -0,0 +1,2 @@
+projectName=ktx-textratypist
+projectDesc=TextraTypist widgets for LibGdx applications.
```
czyzby: `libGDX`.

textratypist/src/main/kotlin/ktx/textratypist/factory.kt
```diff
@@ -0,0 +1,56 @@
...
+import com.github.tommyettinger.textra.TypingLabel
+import ktx.scene2d.KWidget
+import ktx.scene2d.Scene2DSkin
+import ktx.scene2d.Scene2dDsl
+import ktx.scene2d.actor
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.InvocationKind
+import kotlin.contracts.contract
+
+const val DEFAULT_STYLE = "default"
```
czyzby: Please import `ktx.scene2d.defaultStyle` instead.

textratypist/src/test/kotlin/ktx/textratypist/factoryTest.kt
```diff
@@ -0,0 +1,149 @@
...
+
+import com.badlogic.gdx.scenes.scene2d.Actor
+import com.badlogic.gdx.utils.Align
+import com.github.tommyettinger.textra.TextraLabel
+import com.github.tommyettinger.textra.TypingLabel
+import ktx.scene2d.*
+import org.junit.Assert.*
+import org.junit.Test
+
+class BaseFactoryTest : ApplicationTest() {
```
czyzby, line 23: You can import the test sources of the `scene2d` project specifically to have access to the test utilities from there. Please check the Gradle config and unit tests of the `vis` project.

textratypist/src/test/kotlin/ktx/textratypist/testUtilities.kt
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Utility value for numeric tests.
+ */
+const val TOLERANCE = 0.0001f
+
+/**
+ * Tests that require mocked libGDX environment should inherit from this class.
+ */
+abstract class ApplicationTest {
```
czyzby: Please remove this class and import the `scene2d` test module instead, check the previous comment.

textratypist/src/test/resources/test-skin.atlas
```diff
@@ -0,0 +1,20 @@
+
+test-skin.png
+size: 512, 512
+format: RGBA8888
+filter: Linear, Linear
+repeat: none
+Roboto-Black
```
czyzby, line 7: What's the license on the font? I remember that libGDX used to ship with a font and it was possible to reference it in atlases, would that be an option?

czyzby (changes requested):
Looks pretty good, most comments are related to the test setup and documentation rather than the actual library sources. That being said, the support for the library is pretty basic - I originally thought that we could design and maintain a DSL for the markup syntax itself. Something along these lines:

```kotlin
richText {
   red()
   +"This text is red! "
   clearColor()
   +"Now it's not."
}
```

I know that's a much larger undertaking, so if it's not something you'd want to work on, don't worry about it. I'm happy to merge this as is either way.

Anyway, I'd appreciate it if you looked into the default libGDX font or using the VisUI assets to avoid putting a skin atlas and the associated image in test resources before we merge this.

textratypist/README.md
czyzby: I'd pretty much include all major types of markup here for the documentation to be "good enough" without looking up TextraTypist docs.

michaeloa:
> Looks pretty good, most comments are related to the test setup and documentation rather than the actual library sources. That being said, the support for the library is pretty basic - I originally thought that we could design and maintain a DSL for the markup syntax itself.

I'm not averse to adding more stuff (I've customized my own code a fair bit, but mostly with handling styles, etc), but not sure I see the utility of a function DSL for formatting. After all, everything here is just strings.

`richText { red()+"This text is red! "+clearColor()+"Now it's not." }`

Would be:

`textraLabel("[RED]This text is red! []Now it's not.")` 

I would argue that very little is gained by creating `fun red() = "[RED]"`. 

I'd almost say the opposite - `[/]` is a lot less characters than `italic()` - coding `[]` is a lot faster than `clearColor()`. To me, it would be a lot more useful to have some vals for frequently used styles - e.g., `const val RED = "[RED]"` - I've done that myself in my current project, but even for that I'm not convinced it makes sense to make that part of a library as opposed to the user simply defining the exact few such constants they require for their own project.

That is at least my thoughts based on my own usage, I feel like such an addition would obfuscate more than simplify usage. Not to mention that even within textratypist there are some differences between typinglabel and textralabel that I suspect would cause confusion.

textratypist/README.md
michaeloa: Will do. I can certainly add some more markup examples, these were just a few off the top of my head. Not sure where I'd draw the line at "good enough", though... TextraTypist supports a pretty insane number of effects and styles.

{} does not stop animations, btw (that just outputs `{}`), one could use {ENDWAVE} or {RESET} (which [] is just the short form of).

textratypist/src/test/resources/test-skin.atlas
michaeloa: Roboto is OFL (Open Font License), so requires very little (though yeah - should include the license). I can look into the suggested alternative.

textratypist/src/test/kotlin/ktx/textratypist/testUtilities.kt
michaeloa: Hmm. TextraTypist requires FWSkin (or some other variant like FreeTypist) to provide the styles required for textratypis to work correctly though. It doesn't work with your default test application class, because the VisUI provides styles are not compatible. Though maybe I can get away with just replacing the skin loading. Will see what I can do.

michaeloa: Though trying to figure that out reminds me why that it's been almost a decade since I did any substantial programming with vanilla JUnit. Trying to remember how it works isn't easy. 😁

michaeloa:
Ugh. Sorry about some of those changes - I've the ktlint checker auto-enabled in my IntelliJ, and it tends to go crazy about files that don't conform, even when one don't intent to change them. Will try to see if I can reset the pointless changes.

czyzby:
If we're looking strictly at the number of characters used, sure, there's not much to gain. However, the main goals of the DSL would be the compile-time checks and discoverability of the operations. A typo in a string template could result in runtime errors or invalid rendering that could be difficult to pick up or even validate with unit tests. DSL would mostly get rid of these errors by validating the syntax at compile time on top of providing code completion and listing of the supported tags.

Like I said though, it's not critical to add a DSL to get this merged. You're welcome to try if you'd like to design one, but I think there's value in this module as is.

czyzby:
By the way, the syntax I came up with is perhaps not the best for dealing with strings. I like your idea of overloading `+` or perhaps even doing a fluid API. Something like this could be interesting:

```kotlin
richText { red("This is red!") + " This is not." }
```

michaeloa:
Having a busy week, so not much time to fix the last few issues here. Will hopefully find some time in the weekend.

czyzby:
No worries, take your time. Thanks for looking into this!

textratypist/src/test/resources/test-skin.atlas
michaeloa: Added the license.

TextraTypist requires loading a FWSkin in order to create the style classes used by it's widgets. Looked into a few ways to handle this, but of the options I found, this still seem like the least fragile/complicated.

michaeloa:
Think this is ready now.

Ktx and Textratypist are my two go-to libs for game development, so I'll try to set some time aside to add more later.

czyzby (approved):
Thanks, this looks great. I'll try to sort out KTX dependencies next week and do a proper libGDX version bump so that we can use the latest TextraTypist release.

buildSrc/src/main/kotlin/ktx/Versions.kt
```diff
@@ -7,6 +7,7 @@ const val artemisOdbVersion = "2.3.0"
 const val ashleyVersion = "1.7.4"
 const val gdxAiVersion = "1.8.2"
 const val visUiVersion = "1.5.5"
+const val textraTypistVersion = "2.1.9" // Cannot upgrade to 2.2.x until gdx 1.14.x.
```
tommyettinger: There's 2.1.11 out now (or 2.2.10 if you can use libGDX 1.14.0) and I just pushed a small bugfix release today, again! I've been backporting releases to 1.13.1 mostly so KTX users can still use TextraTypist. It isn't that hard to backport, I have it understood well now and it's pretty much one find+replace.

PR #508, opened 2025-11-29
Update to LibGDX 1.14.0 and add PointMapObject to tiled shape extension

Quillraven:
With LibGDX 1.14.0 we also have a new PointMapObject which I added to the shape extension function. It also contains some breaking changes regarding the different tiled map loader parameter classes because they were all empty and redundant. So it got reduced to a single "base" parameter class.

I noticed that VisUI tests are failing because the getText method that returns a StringBuilder doesn't exist. StringBuilder got replaced with CharArray from what I can tell. However, Vis-UI hasn't rolled out the new version yet that supports LibGDX 1.14.0. I see there are commits that fix the mentioned issue but the version was not released yet.
Therefore, I added @Ignored to the failing tests for now.


czyzby:
Thanks, I'll do a proper release hopefully next week - I'll try to get the VisUI out too. I want to push the 1.13.5 one too for consistency.

